PR #11, opened 2024-02-09
feat(cram): slice local files

AssafSternberg:
I have added a slice() and a slice_cram() functions.

MODO.slice(data: str, coords: str) will call slice(), which in turn will call the right slicing function depending on the data format/type that is being sliced (e.g., if it is a CRAM file, it will call slice_cram())

The two things missing ... please add them in ... are, 

(1) determining the data_type (CRAM/array/etc.) of the data, so the right slicing function can be called.

(2) determining the data_path, so the slicing function will know where the data is, and whether it is a local file or on a server (logic for this will be added in later, currently it deals with local files).


modo/cram.py
```diff
@@ -3,16 +3,51 @@
...
+    -----
+    >>> x = slice("data/ex1/demo1.cram", "chr1:100-200")
+    >>> for read in x:
+            print(read)
     """
-    # https://htsget.readthedocs.io/en/stable/index.html
-    ...
+
+    # split up coordinate string "chr:start-end" into its three elements
+    coords = coords.replace("-", ":")
```
cmdoret, line 36: **fix:** this is dangerous, as it is possible for a chromosome name to contain `-`. This would break in such cases.

modo/cram.py
```diff
@@ -3,16 +3,51 @@
...
+
+
+def slice_cram(cram_path: AlignmentFile, coords: str):  # -> AlignmentFile:
+    """Return a slice of the CRAM File as an iterator object.
+
+    Usage:
+    -----
+    >>> x = slice("data/ex1/demo1.cram", "chr1:100-200")
+    >>> for read in x:
+            print(read)
```
cmdoret, line 32:
```suggestion
```

**docs:** this usage describes the `slice` function.

cmdoret (changes requested):
:+1:  I've made a few suggestions and comments and will apply a few commits to address them.

modo/cram.py
AssafSternberg: This is left over from before the split into slice() and slice_cram(), and I forgot to change this to slice_cram()

PR #17, opened 2024-02-26
Refactor/hierarchy

almutlue:
## Main changes in this PR:

1. New modo hierachy with only one level of depth using the `element_type.name`
2. Using zarr path as `element.id`, e.g. `sample/xyz`. Fail early if element with existing id is attempted to be added, repeat prompt with an error message for cli.
__Obs:__ Duplicated id stems are not allowed with the same modo, even though they could be unique by having different zarr paths.
3. File handling when building modod from yaml file. All paths specified in `data_path`will be copied into modo (same behaviour as `modo add -f path/to/some/file MODO_PATH data`). 
__Obs:__ The user can?t just specify a `data_path` without copying the file and this will fail if `data_path` does not exist. This is possible via cli using `modo add MODO_PATH data`.

#### Open Questions:
Currently, we can not add elements of a type that is not specified in `modo.helpers.ElementType`, e.g. `ReferenceSequence`.
Should those go into the main 4 categories: data, sample, assay, references? It would be neat to simply add them to the `ElementType` class (ReferenceSequence= reference), but this will break the internal logic of ElemetType?


cmdoret:
> Should those go into the main 4 categories: data, sample, assay, references?

I think `ElementType` is meant to expose types that we want to publicly expose to the user for modification.
Internally, we could manipulate other types automatically. For example, a user may explicitely  `add` a `ReferenceGenome`, or an `AlignmentFile`, and we would automatically add the associated `ReferenceSequence` behind the scenes. Do you see potentiial issues with this ?

modo/file_utils.py
```diff
@@ -3,26 +3,20 @@
...
         case "CRAM":
-            reference = (
-                instance.has_reference[0]
-                if len(instance.has_reference) == 1
-                else None
-            )
-            cramfile = AlignmentFile(
-                instance.data_path, mode="rc", reference_filename=reference
-            )
+            cramfile = AlignmentFile(modo_path / instance.data_path, mode="rc")
```
cmdoret, line 19: **suggestion(non-blocking):** modo_path -> base_path ?

cmdoret (commented):
Nice! :smiley:  I had a few suggestions, but mostly nitpicks. One thing I notice is that we keep recomputing the list of ids. At some point (maybe in another PR) we may want to have a helper method/function for better clarity (and speed if we add caching).

modo/io.py
```diff
@@ -59,19 +59,33 @@ def parse_multiple_instances(path: Path) -> List:
...
         )
-    # TODO: Could we rename id_ to id?
     modo_dict = modo_inst[0]._as_dict
     modo = MODO(path=object_directory, **modo_dict)
     for instance in instances:
         if not isinstance(instance, model.MODO):
-            modo.add_element(instance)
+            if (
+                isinstance(instance, model.DataEntity)
+                and not modo.path in Path(instance.data_path).parents
```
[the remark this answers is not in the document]
almutlue: yes, I'm sure, but I think you are right and currently that is not a real use case, as the modo will usually be created with `build_modo_from_file()` and thus the path should not exist within modo. I thought this could be useful if we also allow updating modo's via yaml file, but currently that is anyway not implemented/working. Happy to remove the check?

almutlue:
Thanks a lot @cmdoret for all the suggestions and comments! 
I agree with you about the `ElementType` and I think it is plausible to keep this for publicly exposed types.  
I'm not quite sure how to translate this to the `modo.add_element()`. Should this also be scoped to publicly exposed types only (current, see below) or shall we extend its scope? 

```py
def add_element(
        self,
        element: model.DataEntity | model.Sample | model.Assay,
        data_file: Optional[Path] = None,
        part_of: Optional[str] = None,
    ):
       ...
        # Inferred from type inferred from type
        type_name = ElementType.from_object(element).value
        type_group = self.archive[type_name]
        element_path = f"{type_name} / {element.id}"
       ...
```
I see advantages for both. It is nice to have a clear scope for potential api user, but currently we are using `add_element` for all types, which simplifies maintenance?

cmdoret:
> Thanks a lot @cmdoret for all the suggestions and comments! I agree with you about the `ElementType` and I think it is plausible to keep this for publicly exposed types. I'm not quite sure how to translate this to the `modo.add_element()`. Should this also be scoped to publicly exposed types only (current, see below) or shall we extend its scope?
> 
> ```python
> def add_element(
>         self,
>         element: model.DataEntity | model.Sample | model.Assay,
>         data_file: Optional[Path] = None,
>         part_of: Optional[str] = None,
>     ):
>        ...
>         # Inferred from type inferred from type
>         type_name = ElementType.from_object(element).value
>         type_group = self.archive[type_name]
>         element_path = f"{type_name} / {element.id}"
>        ...
> ```
> 
> I see advantages for both. It is nice to have a clear scope for potential api user, but currently we are using `add_element` for all types, which simplifies maintenance?

As discussed, let's have two sets of types:
`ElementTypes`: all supported types in the MODO.
`UserElementTypes`: types that are exposed to the user for editing. (the current `ElementTypes`)

modo/api.py
```diff
@@ -161,53 +178,87 @@ def remove_element(self, element_id: str):
...
-                    f"Cannot make {element.id} part of {part_of}: {parent_type} does not have property {has_prop}"
-                )
-            # has_part is multivalued
-            if has_prop not in self.archive[part_of].attrs:
-                self.archive[part_of].attrs[has_prop] = []
-            self.archive[part_of].attrs[has_prop] += [element_path]
+
+        # Copy data file to archive and update data_path in metadata
+        if data_file is not None:
+            data_path = Path(data_file)
```
cmdoret, line 247:
```suggestion
```
note: I think we don't use `data_path` at all

modo/helpers.py
```diff
@@ -23,13 +24,79 @@ def dict_to_instance(element: Mapping[str, Any]) -> Any:
     )
 
 
+def set_part_of_relationship(
```
cmdoret, line 27:
```suggestion
def set_haspart_relationship(
```
**suggestion:** for consistency, we may want to use `haspart` instead of `partof`, as this matches what the properties we're adding (`has_sample`, `has_assay`, ...)

modo/helpers.py
```diff
@@ -23,13 +24,79 @@ def dict_to_instance(element: Mapping[str, Any]) -> Any:
...
+    element: model.DataEntity
```
cmdoret, line 28: **question:** do we really need to pass `element` around? we use `element.id` and `element.__class__.__name__`.

**comment:** I think `id` could be extracted / replaved by `element_path`. For the class name, we could either use the string like in `get_haspart_property`, or use an enum in both function if that make sense.

cmdoret: An example of what I mean:
```py
def set_haspart_relationship(
    child_class: ElementType | UserElementType,
    child_path: str,
    parent_group: zarr.hierarchy.Group
):
    ...
def get_haspart_relationship(child_class: ElementType | UserElementType):
    ...
```

Or alternatively:
```py
def set_haspart_relationship(
    child_class: str,
    child_path: str,
    parent_group: zarr.hierarchy.Group
):
    ...
def get_haspart_relationship(child_class: str):
    ...
```

modo/helpers.py
```diff
@@ -23,13 +24,79 @@ def dict_to_instance(element: Mapping[str, Any]) -> Any:
...
+
 class ElementType(str, Enum):
-    """Enumeration of element types."""
+    """Enumeration of all element types."""
 
     SAMPLE = "sample"
     ASSAY = "assay"
     DATA_ENTITY = "data"
     REFERENCE_GENOME = "reference"
+    REFERENCE_SEQUENCE = "reference"
```
cmdoret, line 99: **note:** I think it could cause issues down the line if two variants have the same value.
**suggestion:** maybe something like `sequence`, or `reference_sequence`?

almutlue: yes, that is a fair point. I was a bit unsure about this. How do we do with the `Elementtype.from_object()` function? 
Do we dosomething like:
```py
@classmethod
    def from_object(cls, obj):
        """Return the element type from an object."""
        if isinstance(obj, model.Sample):
            return ElementType.SAMPLE
       ...
        elif isinstance(obj, model.ReferenceSequence):
            return "reference"
        else:
            raise ValueError(f"Unknown object type: {type(obj)}")
```
Or is there a more elegant way to solve this? We could also go with a modo structure including an "own" group for sequence like this? :
```bash
/
 ├── assay
 │   └── assay1
 ├── data
 │   └── demo1
 ├── reference
 │   └── reference1
 ├── sequence
 │   └── seq1
 └── sample
     └── sample1
```

almutlue: I chose the second approach in https://github.com/sdsc-ordes/modo-api/pull/17/commits/7f6bcad1462b4933e534baa570fe562a4ea46030 with a separate `sequence` zarr group, because it allows a coherent use of `ElementType` without adding extra logic for the reference sequences. 
Let me know if you think we should only expose the groups `UserElementTypes` on the zarr tree and add ReferenceSequences to the reference group instead? I'm really not sure about this and see advantages for both sides. Adaptations were minor and we can easily rollback or readjust? 

Other adaptations in https://github.com/sdsc-ordes/modo-api/pull/17/commits/62e476f89ee2bfa4d01c1be36ab853f7fe8b435a and https://github.com/sdsc-ordes/modo-api/pull/17/commits/315ec9ea777bf3f175b45604bfb03c4299ce03bc are related to adjusting ids in `has_part` relationships to full ids when building modo from yaml file. I realized that they were still using the short ids (`element_name`) instead of full ids (`element_type / element_name`).

cmdoret: Nice!! That makes sense :smiley:

cmdoret (approved):

Issue #15, opened 2024-02-21
instantiation of remote MODO 


cmdoret:
Currently the API allows read-only access to a catalogue of modo. It supports listing them and their metadata.

**Objective:** We need the ability to remotely instantiate a MODO using the S3 bucket as zarr store.

**Requirements:**

* Client side:
  * [ ] MODO supports instantiation from S3 path (should already work via the `archive` parameter of the constructor)
    + provide a helper function to make this easier?
* Server side:
  * [ ] endpoint should give the **public** address of the bucket to the client

In essence, the client should be able to:
* get the s3 path of a specific modo from the server
* instantiate a modo using that path (as if it was local)

cmdoret:
Here are the proposed functionalities:

Client asks for a list of modo names from server:

```mermaid
sequenceDiagram
    client->>server: list
    server ->>s3: list
    s3 ->> server: modos
    server ->> client: modos
```

Client asks for metadata from all/specific modos on server:

```mermaid
sequenceDiagram
    client->>server: meta
    server ->>s3: request
    s3 ->> server: mount
    server ->> server: read metadata
    server ->> client: metadata
```



Client asks for S3 path of (a) specific modo(s)

```mermaid
sequenceDiagram
    client->>server: "this-modo"
    server ->>client: "http://s3.modo.example.org/modo-bucket/this-modo"
    client ->> client: MODO("http://...this-modo")
```

PR #20, opened 2024-03-06
Feat/remote zarr

almutlue:
Related to https://github.com/sdsc-ordes/modo-api/issues/15:
__Main goal:__
Add capabilities to locally instantiate remote modo(s).

__Includes__:
- `modo.api.MODO()` accepts and handles remote s3 paths
- modo.remote includes new functions `list_remote_items()`, `get_metadata_from_remote()`  and `get_s3_path()` to directly access modo.server endpoints
- new modo.server endpoint to get the public s3 path of all modos match a query

__Open question__:
1. I'm happy to add matching cli functions e.g. `modo list REMOTE_URL` etc.. Would that be useful?
2. `modo.helpers.split_s3_url()` works with simple and unstable string manipulation. I would prefer something more robust! If we want to keep the current setup with one remote path we need to somehow split it into `s3 endpoint url` and `bucket path + modo_path`to align with the syntax of `zarr.open_gorup()`. I couldn't find a good alternative to this. We could do something like: 
```py
class MODO:
    def __init__(
        self,
        path: Union[Path, str],
        s3_endpoint: Optional[str]
    ):
        if s3_endpoint:
            archive=...
``` 
This would require the user to specify the s3_endpoint in addition, though.
Any preferences or ideas to better solve this issue?

### DIFF
--- a/deploy/modo-server/server.py
+++ b/deploy/modo-server/server.py
@@ -18,6 +18,7 @@ import zarr
 
 
 S3_LOCAL_URL = os.environ["S3_LOCAL_URL"]
+S3_PUBLIC_URL = os.environ["S3_PUBLIC_URL"]
 BUCKET = os.environ["S3_BUCKET"]
 HTSGET_LOCAL_URL = os.environ["HTSGET_LOCAL_URL"]
 
@@ -63,4 +64,4 @@ def get_s3_path(query: str, exact_match: bool = False):
             for modo in modos
             if re.search("/" + query + r"$", modo) is not None
         ]
-    return [f"{S3_LOCAL_URL}/{modo}" for modo in res]
+    return [f"{S3_PUBLIC_URL}/{modo}" for modo in res]
--- a/deploy/modo-server/server.py
+++ b/deploy/modo-server/server.py
@@ -11,11 +11,8 @@ import os
 
 from fastapi import FastAPI
 from modo.api import MODO
-<<<<<<< HEAD
-=======
 import rdflib
 import re
->>>>>>> d2468b5 (feat: Add get?query to modo server)
 import s3fs
 import zarr
 
@@ -46,8 +43,9 @@ def gather_metadata():
         archive = zarr.open(
             store=store,
         )
-        # TODO: Fix id_ to id once restructure/herarchy is accepted!
-        meta.update(MODO(path=f"{S3_LOCAL_URL}/{modo}", archive=archive).metadata)
+        meta.update(
+            MODO(path=f"{S3_LOCAL_URL}/{modo}", archive=archive).metadata
+        )
 
     return meta
 
@@ -65,4 +63,4 @@ def get_s3_path(query: str, exact_match: bool = False):
             for modo in modos
             if re.search("/" + query + r"$", modo) is not None
         ]
-    return [f"{S3_URL}/{modo}" for modo in res]
+    return [f"{S3_LOCAL_URL}/{modo}" for modo in res]
--- a/deploy/modo-server/server.py
+++ b/deploy/modo-server/server.py
@@ -48,7 +48,7 @@ def get_s3_path(query: str, exact_match: bool = False):
     modos = minio.ls(BUCKET)
     if exact_match:
         res = [
-            modo for modo in modos if query in modo.replace(BUCKET + "/", "")
+            modo for modo in modos if query in modo.removeprefix(BUCKET)
         ]
     else:
         res = [
--- a/deploy/modo-server/server.py
+++ b/deploy/modo-server/server.py
@@ -41,6 +41,7 @@ def gather_metadata():
         archive = zarr.open(
             store=store,
         )
-        meta = MODO(path=f"{S3_LOCAL_URL}/{modo}", archive=archive).metadata
+        # TODO: Fix id_ to id once restructure/herarchy is accepted!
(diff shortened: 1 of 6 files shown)


deploy/modo-server/server.py
```diff
@@ -11,11 +11,14 @@
 
 from fastapi import FastAPI
 from modo.api import MODO
+import rdflib
```
cmdoret, line 14:
```suggestion
```
**question:** do we need this import?

deploy/modo-server/server.py
```diff
@@ -41,6 +44,24 @@ def gather_metadata():
...
+    modos = minio.ls(BUCKET)
+    if not exact_match:
+        res = [modo for modo in modos if query in modo]
+    else:
+        # NOTE: Is there a better/more roboust way than regexpr?
+        res = [
+            modo
+            for modo in modos
+            if re.search("/" + query + r"$", modo) is not None
+        ]
```
cmdoret, line 66: **suggestion:** just found out about [`difflib.SequenceMatcher()`](https://docs.python.org/3/library/difflib.html#sequencematcher-objects), which is in python's standard library:

```python
>>> import difflib
>>> difflib.SequenceMatcher(None, 'example', 'example').ratio()
1.0
>>> difflib.SequenceMatcher(None, 'example', 'exemple').ratio()
0.8571428571428571
>>> difflib.SequenceMatcher(None, 'example', 'many examples').ratio()
0.7
>>> difflib.SequenceMatcher(None, 'lion', 'zebra').ratio()
0.0
```
We could maybe just apply a threshold on it?
There's also a `.real_quick_ratio()` which uses a heuristic to go much faster. We could also use this one?

cmdoret:
```python
>>> %timeit difflib.SequenceMatcher(None, 'the lazy dog jumps over the lazy dog', 'the quick brown fox jumps over the lazy dog').ratio()
82.7 µs ± 5.29 µs per loop (mean ± std. dev. of 7 runs, 10,000 loops each)

>>> %timeit difflib.SequenceMatcher(None, 'the lazy dog jumps over the lazy dog', 'the quick brown fox jumps over the lazy dog').real_quick_ratio()
9.83 µs ± 603 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

modo/api.py
```diff
@@ -57,7 +59,15 @@ def __init__(
...
     ):
-        self.path: Path = Path(path)
+        if is_uri(path):
+            s3_endpoint, modo_path = split_s3_url(path)
+            archive = zarr.open_group(
+                f"s3://{modo_path}/data.zarr",
+                storage_options={"anon": True, "endpoint_url": s3_endpoint},
+            )
+        else:
+            self.path = Path(path)
```
cmdoret, line 69: **praise:** neat! I didn't realize you could do it so easily with zarr :)
**question:** this is simpler than what we use in modo_server (below). Would it make sense to  use the same pattern in there? Could we even remove the `archive` of `MODO` altogether?

https://github.com/sdsc-ordes/modo-api/blob/46ce93862bd6caab73d1ec65c89fab8735c1b7f6/deploy/modo-server/server.py#L42-L46

modo/helpers.py
```diff
@@ -186,6 +186,15 @@ def is_uri(text: str):
         return False
 
 
+def split_s3_url(s3_url: str) -> tuple[str, str]:
```
cmdoret, line 189: **note:** this does indeed feel dangerous indeed if we keep that, I'd start by checking if the format matches our assumption (if possible) and raise an error if it doesn't.

modo/remote.py
```diff
@@ -0,0 +1,52 @@
...
+import requests
+from typing import List, Mapping, Optional
+
+
+def list_remote_items(remote_url: HttpUrl) -> List[HttpUrl]:
+    return requests.get(url=remote_url + "/list").json()
+
+
+def get_metadata_from_remote(
+    remote_url: HttpUrl, id: Optional[str] = None
```
cmdoret, line 13: **praise:** simple yet very useful :D
**suggestion:** avoid shadowing built-in `id` (`id_`, `modo_id`, ...)

modo/remote.py
```diff
@@ -0,0 +1,52 @@
...
+            raise ValueError(
+                f"Could not find metadata for modo with id: {id}"
+            ) from e
+    else:
+        return meta
+
+
+def get_s3_path(
+    remote_url: HttpUrl, query: str, exact_match: bool = False
+) -> List:
```
cmdoret, line 38:
```suggestion
) -> list[str]:
```
**note:** we require python >= 3.10, so lowercase type hints are legal :)

cmdoret (commented):
This adds very useful functionality and the code is surprisingly simple! :100:

 I agree that the current implementation of `split_s3_url` is brittle and like your suggestion of having a separate `s3_endpoint` argument. This feels like a safer solution!

cmdoret (changes requested):
Just had one suggestion for for a safer path processing when removing the bucket, otherwise everything looks great!

the changes are very minor and I'll apply them directly, hope that's ok

cmdoret:
I took the liberty to apply my suggestions for safer bucket replacement and added a bit of code to sort the output by similarity,

Since there was nothing else for me, I'll approve the PR and let you review the change I made.

cmdoret (approved):

almutlue:
Thanks a lot! Looks really good to me. I didn't know `removeprefix()` :-)

Issue #29, opened 2024-03-21
Code clarity: Restructure code dividing modo into remote and local subclass


almutlue:
Currently, we use __conditionals__ to enable __remote__ and __local__ use of the modo class and it's functions. 
Instead, we should use an abstract class and child classes, similar as gimie extractors in https://github.com/sdsc-ordes/gimie/blob/main/gimie/extractors/abstract.py that manage the differences in code. 
This would help to keep class functions clean and simple and enable sequential addition of new features.

PR #30, opened 2024-03-21
Feat/remote add

almutlue:
#### 1. Bug fix to remove zarr element group itself in `modo.element.remove(element_id)`:
__Ops:__ this fix is unrelated to remote writing, but is a bug detected during the PR.
Before `modo.element.remove(element_id)` removed data files associated to the element that should be deleted and all links to other elements, but not the element's zarr group itself. 

#### 2. Fix `modo.list_arrays()` to correctly display file tree for local and remote objects:
`list_arrays()` calls the `zarr.group.tree()` function. For remote objects this functions does not use consollidated metadata (even, if `zarr.consollidate_metadata()` was called before) resulting in file trees with duplicated elements. Explicitly opening the consollidated zarr store fixes this issue __and__ speeds up the process noteably!

#### 3. Use s3fs filestore to handle remote file objects:
Remote files that are not written in zarr need to be accessed and handled via s3fs.  To add this all file handling functions had to be adapted by including conditionals on the filestore (local or remote). This affected api functions related to adding, removing and listing files as well as some of there internal functions.
__Ops:__: This resulted in multiple conditionals. We should restructure the code to use child classes managing the differences (see https://github.com/sdsc-ordes/modo-api/issues/29) 

#### 4. Add remote capabilities to cli:
So far cli functions did not comply with remote objects. By adding a new cli-argument (`--s3-endpint`, `-s3`) they can automatically use the modo.api remote capabilities. 

#### 5.Fix remote modo creation
 As previously remote modos were read only, we did not differentiate between existing and non-existing modos in `MODO()` . To allow creating new modos remotely `MODO()` can now distinct between remote/local and existing/non-existing for each of them.


modo/api.py
```diff
@@ -125,17 +134,28 @@ def show_contents(self):
...
-            elif path.is_file():
-                yield path
-            for file in path.rglob("*"):
-                yield file
+        if isinstance(self.archive.store, zarr.storage.FSStore):
+            fs = self.archive.store.fs
+            for path in fs.find(str(self.path)):
+                if Path(path).name.endswith(".zarr") or Path(
+                    path
+                ).name.startswith("."):
```
supermaxiste, line 142: **clarification**: is there a reason why we're excluding dot files on the remote and not on local?

modo/api.py
```diff
@@ -170,6 +190,13 @@ def remove_element(self, element_id: str):
             data_file = self.path / attrs["data_path"]
             if data_file.exists():
                 data_file.unlink()
+            elif isinstance(
+                self.archive.store, zarr.storage.FSStore
+            ) and self.archive.store.fs.exists(data_file):
+                self.archive.store.fs.rm(str(data_file))
```
supermaxiste, line 196: **suggestion**: removal of files happens silently, would it make sense to have some messages to either confirm that a file was deleted or found? I guess the only way currently is to check the bucket directly, so a message could help.

modo/api.py
supermaxiste, line 1: **clarification/suggestion**: the functions `add_element` and `_add_any_element` look almost identical to me, except for sequences being allowed in one and not the other. Isn't there a way of consolidating them?

supermaxiste (commented):
Excellent job @almutlue!

**praise**: code readability is excellent and your changes are very understandable even by just looking at the code.

My comments are mostly when it comes to code itself. I'm also leaving some small issues I found when using `modo`:

- **enhancement (non-blocking)**: when creating a new object, I typed by mistake a character string on the field `taxon_id`. `modo` threw out a type error only when I finished filling all the fields. It would be great to have a check on the spot and/or state when a certain type is expected.
- **bug/enhancement**: when creating a remote object that has an ID that already exists on the remote, `modo` will overwrite the object without letting you know there was one there already. 
- **enhancement(non-blocking)**: `modo --help` displays only part of the description for some commands (see `publish` and `remove` below). 
```shell
(modo-py3.10) (base) stefan@ordesmain:~/modo-api$ modo --help
Usage: modo [OPTIONS] COMMAND [ARGS]...

Options:
  --help  Show this message and exit.

Commands:
  add      Add elements to a digital object.
  create   Create a digital object.
  publish  Creates a semantic artifact allowing to publish a digital...
  remove   Removes the target element from the digital object, along with...
  show     Show the contents of a digital object.
```

modo/api.py
almutlue: Thanks for your comment, that is a fair point. @cmdoret and I had the discussion as well. The idea is that `add_element` is the user function with limited capabilities and `_add_any_element()` is the internal functions that allows adding the full scope of elements. It means some extra/duplicated code, but gives us better control. Currently, this only affects elements from the `Sequence` class, but I think this could potentially be extended in future depending on the data model. So the intention is to make a clear distinction between what a user should add and what we use internally. Maybe we can keep it like this for now and revisit in some time (after we got some user feedback) to see if this is really necessary and useful?

modo/api.py
almutlue: Very good point and well spotted! The reason was in the different return of `fs.find(self.path)` (returns all files at file level) and `self.path.glob("*")` (returns the files and subdirectories of path). So this part `Path(path).name.endswith(".zarr")` did not properly exclude the zarr archive. Excluding dot files was a dirty fix and could indeed result in a different behavior between local and remote listing. In 1ab950d699cf0756bb89e019fa56da187c611281 I changed the conditionals to explicitly exclude the zarr archive and list all files and related files in subdirectories.

almutlue:
> bug/enhancement: when creating a remote object that has an ID that already exists on the remote, modo will overwrite the object without letting you know there was one there already.

Very good point! The behavior was indeed different from local creation and non-transparent.

**cli**: In 6e1e0afc2646c7f1ed8e5e5bda1fa6e9c63c61a6 I added a check to raise a ValueError if the `object_directory` exists at the remote storage to `modo create`.

**api**: The same command (`MODO()`) can be used to create a new or load an existing object, depending on the path existance. To be more transparent, in 6e1e0afc2646c7f1ed8e5e5bda1fa6e9c63c61a6 I added a message, if an existing modo was loaded from local or remote:
`f"Load existing MODO from remote path: {path}\n"
 "NOTE: Additional arguments will be ignored."`

modo/api.py
almutlue: Fair point, added a message (after deletion) to **modo.api** for clarification in 1ab950d699cf0756bb89e019fa56da187c611281 and a prompt asking for confirmation to the **cli** in 1f4f3e94d8641d4404636be7591565cd3fef7e4a

almutlue:
Thanks a lot @supermaxiste for the review! You raised very good and useful points. I  tried to address all of them, but the 2 `non-blocking` `enhancement` points. For those I opened issues #34 and #33 and I will look into them in future.

supermaxiste (approved):
Very well done @almutlue, all good on my side 🥂

PR #31, opened 2024-03-28
Feat/cram stream

AssafSternberg:
#### 1. Refactor of the region parser to allow a wider range of range request formats:
The old region parser did not allow the user to ask for an entire chromosome (e.g., "chr1"), a one side open-ended region (e.g., "chr1:10000" or "chr1:-10500"), or un-mapped reads (e.g., "*"). The refactored parser deals with such requests. Due to the possibility of start and/or end to be None, the output was changed to tuple[chrom: str, start: str, end: str].

#### 2. Refactor slice_cram() to handle the updated output of parse_region():
The old version of slice_cram() worked only with a "reference:start-end" region format, expecting start and end as integers. The New version handles start and end being strings or None.

#### 3. Add slice_remote_cram() to handle remote cram files:
The new function streams or writes to a local file the region of a remote CRAM file. The function uses the helper functions:
- filename_from_url(): extracts the filename from the given URL.
- hstget_command(): formats a valid htsget command line request.

#### 4. Add htsget_url function:
Currently not used, but might be useful in the future. Similar to htsget_command(), it returns a valid htsget URL that can be used in command line use (e.g., 'samtools view -H http://localhost/htsget/reads/ex/demo1?referenceName=chr1&start=10000&end=10500&format=CRAM')

#### 5. Add stream_cram method to MODO
The method format is identical for local and remote files. First, it checks that the CRAM file actually exists in the MODO. If it does, based on the existence of s3_endpoint it either calls slice_cram() or slice_remote_cram(). Currently local files are always streamed. For remote files, if an output filename is given, the data is written to a local file. If not, the data is streamed.


modo/helpers.py
```diff
@@ -186,24 +186,111 @@ def is_uri(text: str):
...
+        htsgetURL += f"referenceName={chrom}"
+        if start != "":
+            htsgetURL += f"&start={start}"
+        if end != "":
+            htsgetURL += f"&end={end}"
+        htsgetURL += "&"
+    htsgetURL += f"format={file_extension.upper()}"
 
-    return (chrom, int(start), int(end))
+    return htsgetURL
```
almutlue, line 296: **Suggestion**: Can we use the htseq-api instead of their cli (see comment above). This would enable us to use the api arguments instead of string parsing.

almutlue (commented):
Thanks a lot for the PR. I left some questions and suggestions on mainly minor things, but I have one main point:
- using an `os.system` call to run htseq makes the code more complicated and inconsistent. I think we should use python instead, as this is what the htseq cli calls anyways. We could either use the htseq-api with `output= sys.stdout.buffer`(as they do in https://github.com/ga4gh/htsget/blob/ac3ba43fd017fc1db173d323a5de2987a354fdb6/htsget/cli.py#L72) or simply write our own requests? I did not manage to fully unravel their code, but it is rather short and looks like little extra to the request call. The code looks slightly smelly to me (not addressed todos from >2 years, hard coded values, https://github.com/ga4gh/htsget/blob/master/htsget/io.py). So maybe it is worth to implement by ourselves?

modo/api.py
```diff
@@ -321,3 +323,33 @@ def enrich_metadata(self):
...
+            if cram_name == str(filepath).split("/")[-1]:
+                path = filepath
+                break
+        if path == "":
+            raise ValueError(f"{cram_name} not fount in {self.path}.")
+
+        if self.s3_endpoint:
+            # http://domain/s3 + bucket/modo/file.cram --> http://domain/htsget/reads/modo/file.cram
+            url = (
+                re.sub(r"s3$", "", self.s3_endpoint)
```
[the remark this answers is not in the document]
cmdoret: The user should have the option to specify a custom s3/htsget url. If not specified, we'll infer via substitution (with warning msg maybe). Would that work? @AssafSternberg

modo/api.py
```diff
@@ -321,3 +323,33 @@ def enrich_metadata(self):
...
+
+        # check requested CRAM exists in MODO
+        path = ""
+        filepaths = list(self.list_files())
+        for filepath in filepaths:
+            if cram_name == str(filepath).split("/")[-1]:
+                path = filepath
+                break
+        if path == "":
+            raise ValueError(f"{cram_name} not fount in {self.path}.")
```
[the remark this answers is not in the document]
AssafSternberg: Technically, we could request the user to specify the file path, not only filename. This would make life easier for us, but I think this will not be the best user experience. The way I see it, part of the idea of having MODO is to switch from thinking of the separate data files to thinking of a single object, which the data files are a part of. Therefore, it is more "natural" from the user point of view to specify the MODO object and the data file associated to it, without the need of specifying the file path. I would rather keep things like this, but if both of you feel strongly against it, then we can change this.

modo/api.py
```diff
@@ -321,3 +323,33 @@ def enrich_metadata(self):
...
+            url = (
+                re.sub(r"s3$", "", self.s3_endpoint)
+                + "htsget/reads/"
+                + path.split("/", maxsplit=1)
+            )
+            slice_remote_cram(url, region, output)
+        else:
+            # assuming user did not change directory, filepath should be the
+            # relative path to the file.
+            iter = slice_cram(path, region)
```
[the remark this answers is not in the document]
AssafSternberg: Yes, if we change the above then instead of 'path', we will need to use ;cram_path'.

modo/api.py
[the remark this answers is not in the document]
AssafSternberg: Yes, that should work.

instead of,
modo=MODO(path= 'modo-demo/ex', s3_endpoint = 'http://localhost/s3')

we can use,
modo=MODO(path= 'modo-demo/ex', s3_endpoint = 'http://localhost/s3', htsget_endpoint = 'http://htsget_domain/')

modo/api.py
```diff
@@ -321,3 +323,33 @@ def enrich_metadata(self):
...
+                break
+        if path == "":
+            raise ValueError(f"{cram_name} not fount in {self.path}.")
+
+        if self.s3_endpoint:
+            # http://domain/s3 + bucket/modo/file.cram --> http://domain/htsget/reads/modo/file.cram
+            url = (
+                re.sub(r"s3$", "", self.s3_endpoint)
+                + "htsget/reads/"
+                + path.split("/", maxsplit=1)
```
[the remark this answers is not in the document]
AssafSternberg: Yes, you are correct, I forgot the [1] index.

modo/cram.py
```diff
@@ -10,20 +10,40 @@
...
     return iter
 
 
+def slice_remote_cram(
+    url: str, region: str = None, output_filename: str = None
+):
+    """Stream or write to a local file a slice of a remote CRAM file"""
+
+    htsget_req = htsget_command(url, region, output_filename)
+    os.system(htsget_req)
```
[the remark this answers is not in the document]
AssafSternberg: Yes, now that I can actually use sys.stdout.buffer ... which was not possible via Spyder ... I will change the script to use the htsget api instead of the cli via os.system() .

modo/helpers.py
AssafSternberg: As mentioned in the above comment, yes, I will switch from using the cli to the api.

modo/api.py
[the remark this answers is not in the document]
almutlue: I prefer the path:

1. File names are not neccessary unique, but paths are, e.g. `data/ex/reference.fa` and `data/ex/reference/reference.fa` have the same name, but not the same path. Both path are currently possible in modo.Unless we check for unique names upon adding files this would not be safe. 

2. We return the path in `modo.list_files()`and I think we should be consistent and use one way of referencing the same entity.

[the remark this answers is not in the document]
AssafSternberg: 1. I do indeed think that we need to impose unique file names. Moreover, I think that in real life use, files will tend to be unique and more descriptive (I don;t think "reference.fa" is a good representative of file naming that will be used).

2. That's true, though we could have modo.list_files() that will list the filenames and modo.list_filepaths() that will give the paths.

BTW, one problem with a file path given by the user, is that modo.list_files() gives relative paths, while the user can supply an absolute path, which will need to be handled via string manipulation to trim it to a relative path.

modo/api.py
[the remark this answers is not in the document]
cmdoret: Looks good yes, the default value can be None, and the docstring explain that this means falling back to same domain

modo/api.py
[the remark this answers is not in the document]
cmdoret: Instead of the user providing the filename, how about providing the `name` or `id` of a DataEntity?

e.g. if a user runs `modo show`, there will be a record like:
```yaml
/data/demo1:
  '@type': DataEntity
  data_format: CRAM
  data_path: demo1.cram
  description: Dummy data for tests
  has_reference:
  - reference1
  has_sample:
  - sample1
  name: Demo 1
```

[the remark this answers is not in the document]
cmdoret:
```py
cram_id = 'demo1'

# all cram files are inside `/data` in the modo metadata hierarchy
cram_record = self.metadata[f"/data/{cram_id}"]["data_path"]

if cram_record["data_format"] != 'CRAM':
    raise ValueError("not a cram record")

cram_path = cram_record["data_path"]
```

modo/api.py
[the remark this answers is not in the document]
AssafSternberg: BTW, just noticed that we don't actually initialize 'self.s3_endpoint', rather 's3_endpoint' is just used as a variable. I think we should add 'self.s3_endpoint', and as discussed with Cyril this morning, also a 'self.htsget_endpoint', which can be then used for translating between the s3 path and the htsget path, without needing the users to enter them, and averting problems if they enter wrong or invalid endpoints.

modo/api.py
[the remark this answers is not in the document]
AssafSternberg: Changed stream_cram method to use cram_path instead of cram_name (see commit - https://github.com/sdsc-ordes/modo-api/pull/31/commits/0a1a19183218bf9587148dbf07491ddd71c07427)

`    def stream_cram(
        self, cram_name: str, region: str = None, output: str = None
    ):
        """Slices and streams the requested CRAM file, both local and remote,
        and either outputs a data stream or writes data to local file"""

        # check requested CRAM exists in MODO
        if Path(cram_path) not in self.list_files():
            raise ValueError(f"{cram_path} not found in {self.path}.")

        if self.s3_endpoint:
            # http://domain/s3 + bucket/modo/file.cram --> http://domain/htsget/reads/modo/file.cram
            url = (
                self.htsget_endpoint
                + "/reads/"
                + str(Path(*Path(cram_path).parts[1:]))
            )
            # str(Path(*Path(cram_path).parts[1:])) same as path.split("/", maxsplit=1)[1] but cross-platform
            slice_remote_cram(url, region, output)
        else:
            # assuming user did not change directory, filepath should be the
            # relative path to the file.
            iter = slice_cram(cram_path, region)
            return iter`

modo/api.py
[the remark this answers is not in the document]
AssafSternberg: I added self.s3_endpoint and self.htsget_endpoint to MODO. If not given, self.htsget_endpoint is assumed to have the same domain as s3_endpoint, or None if s3_endpoint is not given as well (see commit - https://github.com/sdsc-ordes/modo-api/pull/31/commits/aa7fc9712fa89a6a5ab3d23457fb552a5ab1c0a5)

`
class MODO:
    """Multi-Omics Digital Object
    A digital archive containing several multi-omics data and records.
    The archive contains:
    * A zarr file, array-based data and metadata pointing to arrays and data files
    * CRAM files, with genomic-alignments data
    Examples
    --------
    >>> demo = MODO("data/ex")
    # List identifiers of samples in the archive
    >>> demo.list_samples()
    ['/sample/sample1']
    # List files in the archive
    >>> files = sorted([file.name for file in demo.list_files()])
    >>> assert 'demo1.cram' in files
    >>> assert 'reference1.fa' in files
    """
    def __init__(
        self,
        path: Union[Path, str],
        s3_endpoint: Optional[str] = None,
        htsget_endpoint: Optional[str] = None,
        id: Optional[str] = None,
        name: Optional[str] = None,
        description: Optional[str] = None,
        creation_date: date = date.today(),
        last_update_date: date = date.today(),
        has_assay: List = [],
        source_uri: Optional[str] = None,
    ):
        self.s3_endpoint = s3_endpoint
        if s3_endpoint and not htsget_endpoint:
            htsget_endpoint = re.sub(r"s3$", "htsget", s3_endpoint)
        self.htsget_endpoint = htsget_endpoint`

modo/helpers.py
AssafSternberg: Switched from htesget CLI to htsget API in slice_remote_cram() (see commit - https://github.com/sdsc-ordes/modo-api/pull/31/commits/833573db60b063778c1901f90a055e74fb97bbd0)

modo/cram.py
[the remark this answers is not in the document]
AssafSternberg: Implemented (see commit - https://github.com/sdsc-ordes/modo-api/pull/31/commits/833573db60b063778c1901f90a055e74fb97bbd0)

AssafSternberg:
Any further comments, @almutlue and @cmdoret , or can I go ahead and merge the branch with main (and if yes, which one should I use 'create a merge commit', 'squash and merge', or 'rebase and merge'?)

modo/api.py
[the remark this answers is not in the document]
AssafSternberg: Fixed, and later refactored to,

` url = (
        self.htsget_endpoint
        + "/reads/"
        + str(Path(*Path(cram_path).parts[1:]))
)`

as it is cross-platform (and won't fail if back-slashes are used)

modo/api.py
```diff
@@ -370,3 +377,28 @@ def enrich_metadata(self):
                     self.update_element(inst_names[ele.name], ele)
                 else:
                     continue
+
+    def stream_cram(
+        self, cram_name: str, region: str = None, output: str = None
```
almutlue, line 382: I think you forgot to change `cram_name` to `cram_path` here?
```suggestion
        self, cram_path: str, region: str = None, output: str = None
```

AssafSternberg: You are correct ... always great to have a second and third set of eyes :)

modo/api.py
```diff
@@ -370,3 +377,28 @@ def enrich_metadata(self):
...
+    ):
+        """Slices and streams the requested CRAM file, both local and remote,
+        and either outputs a data stream or writes data to local file"""
+
+        # check requested CRAM exists in MODO
+        if Path(cram_path) not in self.list_files():
+            raise ValueError(f"{cram_path} not found in {self.path}.")
+
+        if self.s3_endpoint:
```
almutlue, line 391: __Nitpick__(minor): I think it would be more natural to check for `htsget_endpoint` here as we are using it downstream? In principle they should always be paired, so I don't think this would break something, but feels more coherent to me?

AssafSternberg: @cmdoret and I had the same discussion. I agree that checking self.htsget_endpoint in the same as checking self.s3_endpoint, and it won't break anything just to switch from one to the other, but for coherency of reading the code, I think it is more natural to check if there is a self.s3_endpoint, when wanting to check if the file is remote on an s3 bucket.

modo/cram.py
```diff
@@ -10,20 +10,66 @@
...
+        start = int(start)
+    if end == "":
+        end = None
+    else:
+        end = int(end)
+
+    if output_filename:
+        with open(output_filename, "wb") as output:
+            htsget.get(
+                url, output, reference_name, start, end, data_format="cram"
```
almutlue, line 58: **Suggestion(minor)**: With this number of arguments I think it is safer to use named arguments even if they read repetitive:
 ```suggestion
            htsget.get(
                url = url, 
                output = output,
                reference_name = reference_name, 
                start = start, 
                end = end, 
                data_format="cram"
```

File: modo/cram.py [hunk 10]
```diff
@@ -10,20 +10,66 @@
...
+            )
+    else:
+        htsget.get(
+            url,
+            sys.stdout.buffer,
+            reference_name,
+            start,
+            end,
```
almutlue, line 66: same here:
```suggestion
        htsget.get(
            url = url,
            output = sys.stdout.buffer,
            reference_name = reference_name,
            start = start,
            end = end,
```

modo/cram.py
```diff
@@ -10,20 +10,66 @@
...
     cramfile = AlignmentFile(path, "rc")
 
-    iter = cramfile.fetch(chrom, start, stop)
+    iter = cramfile.fetch(chrom, start, end)
 
     return iter
 
 
+def slice_remote_cram(
+    url: str, region: str = None, output_filename: str = None
```
almutlue, line 38: __To-do__: `parse_region()` expects a str, but region is assigned  `None ` as default, which will fail. 
So I think `region` should either be a non-optional argument and we expect the user to specify a string or we use something like "*" as default.

modo/api.py
```diff
@@ -370,3 +377,28 @@ def enrich_metadata(self):
...
+            # http://domain/s3 + bucket/modo/file.cram --> http://domain/htsget/reads/modo/file.cram
+            url = (
+                self.htsget_endpoint
+                + "/reads/"
+                + str(Path(*Path(cram_path).parts[1:]))
+            )
+            # str(Path(*Path(cram_path).parts[1:])) same as path.split("/", maxsplit=1)[1] but cross-platform
+            slice_remote_cram(url, region, output)
```
almutlue, line 399: __Question__: For local slicing we return a `pysam.libcalignmentfile.IteratorRowRegion`, but for remote slicing we return the stream. I think it would be good to have a consistent behavior for remote and local streaming. Is it possible to transform the stream into a similar pysam rowregion as well?

modo/api.py
```diff
@@ -370,3 +377,28 @@ def enrich_metadata(self):
...
+        else:
+            # assuming user did not change directory, filepath should be the
+            # relative path to the file.
+            iter = slice_cram(cram_path, region)
```
almutlue, line 403: __Suggestion__: Similar to the above point, we are not consistent between local and remote behaviour here, but only allow writing the output to an output file for remote streaming. What about adding `output` to local slicing as well?

AssafSternberg: Added 'output_filename' argument (optional) to slice_cram(), to allow the user to copy the slice into a local file (see commit - [bcc16f3a982e8461e43f439195e34db9dc2988a0](https://github.com/sdsc-ordes/modo-api/commit/bcc16f3a982e8461e43f439195e34db9dc2988a0))

AssafSternberg: Added argument and functionality to save_cram() MODO method (see commit - https://github.com/sdsc-ordes/modo-api/pull/31/commits/0b0ed93e2029420d128e1c81b0ce374ec7b128f3)

modo/api.py
AssafSternberg: To harmonize method behavior I split stream_cram() into two methods:
- stream_cram() - streams or returns an iterator (in the future we can further harmonize this)
- save_cram() - saves the output to a local file.
(see commit - https://github.com/sdsc-ordes/modo-api/commit/0b0ed93e2029420d128e1c81b0ce374ec7b128f3)

almutlue (approved):
Thanks Assaf.
This looks good to me and works on my side as well. 
I needed to specify a local reference file to make it work. It would be nice, if it could use the remote reference as default, but I think that is for a new PR.

modo/api.py
AssafSternberg: stream_cram() harmonized as well, returns iterator in both local and remote cases.

modo/cram.py
AssafSternberg: fixed - changed to 'region: Optional[str] = None'

PR #32, opened 2024-04-08
docs: sphinx website with API reference

almutlue:
# Aim: Setup of a basic documentation

## PR elements:
1. Sphinx-based documentation using the [pydata theme](https://pydata-sphinx-theme.readthedocs.io/en/stable/index.html)
2. Automatically generated cli and api docs
3. Docs for background, tutorials, quickstart
4. Landing page linking to all resources + new logo 
5. Github workflow to build docs  
ed
__Ops__: description of htsget slicing and streaming is missing. To be added after https://github.com/sdsc-ordes/modo-api/pull/31 is merged


cmdoret:
> Note: the pytest CI job matrix was running python 3.9 and 3.10, but pyproject.toml restricts to python 3.10+. This caused the ci to run tests twice python 3.10 ([logs](https://github.com/sdsc-ordes/modo-api/pull/32/checks#step:7:17)), bumped ci versions in ae7e373.

.github/workflows/sphinx-docs.yml
```diff
@@ -0,0 +1,64 @@
+name: docs
+on:
+  push:
+    branches: [main]
+  pull_request:
+    paths:
+      - 'docs/**'
```
cmdoret, line 7: **question:** as we use autoapi to build html from python sources, don't we want to rebuild the website on any change impacting the code? (i.e. not only docs directly)

cmdoret (commented):
General:

**nitpick(non-blocking):** just to keep the same convention as the `docs/` folder and `docs` commit label, we could use `docs` instead of `doc` for the makefile rule and the poetry dependency group. (would also need to be updated in the sphinx ci job)

docs/tutorials/modo_arrays.md
```diff
@@ -0,0 +1,156 @@
...
+Finally all arrays can be added to the modo element:
+
+::::{tab-set}
+
+:::{tab-item} python
+:sync: python
+```{code-block} python
+
+modo.archive["data/rna1"].create_dataset("data", data=rna_array)
+modo.archive["data/rna1"].create_dataset("obs", data=obs)
```
cmdoret, line 112: **comment:** I get the error
```
ValueError: missing object_codec for object array
```

Using json as codec seems to fix it (then the array is stored as json on disk):

```py
modo.archive["data/rna1"].create_dataset("var", data=var, object_codec=numcodecs.JSON())
```

docs/tutorials/modo_arrays.md
```diff
@@ -0,0 +1,156 @@
...
+:sync: python
+```{code-block} python
+from modo.api import MODO
+import modo_schema.datamodel as model
+
+# load modo - example at "data/ex"
+modo= MODO("data/ex")
+
+# Generate an Array element
+array_element = model.Array(id="rna1", name= "RNA raw counts", description = "RNA counts from multiple timepoints", has_sample="sample/sample1", data_format = "Zarr")
```
cmdoret, line 79: **comment:** I get 
```
ValueError: data_path must be supplied
```
adding `data_path` with a meaningless value seems to fix it. 

**question**: Should we open a separate issue for this? (i.e. make data_path optional)

cmdoret: alternatively, we could also add a note saying that we plan to add helper functions (an array api) to simplify these operations

docs/tutorials/modo_remote.md
```diff
@@ -0,0 +1,118 @@
...
+#   creation_date: '2024-02-19T00:00:00'
+#   description: Dummy modo for tests.
+#   has_assay:
+```
+:::
+
+::::
+
+:::{note}
+__Ops:__ The __bucket name__ and the __S3 endpoint url__ are specified separatly. The __bucket name__ is prepended to the `MODO`'s name in the same way as a local path, while the __S3 endpoint url__ needs to be specified specifically.
```
cmdoret, line 82: **question:** what does `Ops` mean here?

cmdoret (changes requested):
The docs look beautiful! Incredible work :clap: :

General comments:

**nitpick(non-blocking):** just to keep the same convention as the `docs/` folder and `docs` commit label, we could use `docs` instead of `doc` for the makefile rule and the poetry dependency group. (would also need to be updated in the sphinx ci job)

**comment(non-blocking)** on some pages (e.g. cli), the navigation sidebar on the left is empty. If I understand correctly, this bar is supposed to contain sub-pages. I wondered if the CLI reference page should itself be a sub-page. We can discuss that in a future issue.

docs/conf.py
```diff
@@ -0,0 +1,77 @@
...
+# Options for autoapi
+autoapi_dirs = ["../modo"]
+autoapi_ignore = ["*cli*"]
+autodoc_typehints = "description"
+
+# Options for intersphinx
+
+intersphinx_mapping = {
+    "python": ("https://docs.python.org/", None),
+    "zarr": ("https://zarr.readthedocs.io/en/stable/", None),
```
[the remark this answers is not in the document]
almutlue: Good point and well spotted. 
I think the problem is that zarr imports most of the classes we use internally in `__init__.py`. This has been described [here](https://github.com/sphinx-doc/sphinx/issues/5603) and was [solved for autodoc](https://github.com/sphinx-doc/sphinx/pull/9026), but not yet for [autoapi](https://github.com/readthedocs/sphinx-autoapi?tab=readme-ov-file). There is an [open issue](https://github.com/readthedocs/sphinx-autoapi/issues/411) (last edit in Feb) to resolve this. 
We could:
- switch to autodoc
- wait for the issue to be resolved
- manually add missing imports as described in [this workaround](https://github.com/sphinx-doc/sphinx/issues/5603)
- change our references to the zarr classes exported in `python -m sphinx.ext.intersphinx https://zarr.readthedocs.io/en/stable/objects.inv`

Any preferences?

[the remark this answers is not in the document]
cmdoret: The last option sounds good to me, the types would be a bit more verbose, but I personally don't mind:
AFAICT, we would have to replace the following:
* `zarr.Array` -> `zarr.core.Array`
* `zarr.DirectoryStore` -> `zarr.storage.DirectoryStore`
* `zarr.group` -> `zarr.hierarchy.group`
* `zarr.Group` -> `zarr.hierarchy.Group`
* `zarr.open` -> `zarr.convenience.open`

docs/intro/quickstart.md
```diff
@@ -0,0 +1,68 @@
...
+:sync: pip
+```{code-block} console
+modo create ex
+```
+:::
+
+:::{tab-item} docker
+:sync: docker
+```{code-block} console
+docker run -e ghcr.io/sdsc-ordes/modo-api:latest create ex
```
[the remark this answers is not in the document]
almutlue: Thanks a lot! That was indeed not working. To also allow prompts, I replaced it with:
`docker run -itv "${PWD}:/modo" ghcr.io/sdsc-ordes/modo-api:latest modo create modo/ex`

docs/tutorials/modo_access.md
```diff
@@ -0,0 +1,148 @@
...
+#     modos:omics_type NCIT:C84343 .
+#
+#     ...
+```
+:::
+
+::::
+
+## Enrich metadata
+Genomic files such as cram files store relevant metadata, e.g. their reference sequences, in their header. These information can be automatically extracted and included into a `MODO`.
```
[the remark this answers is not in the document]
almutlue: Currently, there is no option to automatically "enrich metadata" from the cli. I somehow never prioritized it. 
I opened an issue https://github.com/sdsc-ordes/modo-api/issues/37 to not forget it.

cmdoret:
**question:** it would also be nice to have intersphinx mappings to modo_schema, but I'm not sure what works it implies on the modo_schema side. Do we need it to also use sphinx for this?

docs/tutorials/modo_arrays.md
almutlue: Yes, very good point. I missed that - what about doing all of this:
1. In this PR: Fix the docs with the "path to where we will store the array"
2. In this PR: Add the note to make clear this is going to be wrapped/simplified
3. Open an issue to simplify this and make the data path optional (see https://github.com/sdsc-ordes/modo-api/issues/38)

docs/tutorials/modo_arrays.md
almutlue: Thanks a lot, well spotted! I missed that one.

docs/tutorials/modo_arrays.md
cmdoret: Sounds good to me!

docs/tutorials/modo_remote.md
almutlue: Good point, removed it.

almutlue:
Thanks a lot @cmdoret ! I agree with your points and hope I adressed all of them. 
Regarding the missing cli tab: That should affect those where we have not a cli implementation, yet. We should have some issues (https://github.com/sdsc-ordes/modo-api/issues/37 and https://github.com/sdsc-ordes/modo-api/issues/38) pointing to those now

cmdoret (approved):
Great! All the comments are addressed and the interspinx mappings work very nicely :clap:

PR #45, opened 2024-05-21
Add stream and save handling of VCF and BCF files

AssafSternberg:
This PR concerns the addition of VCF and BCF handling, both local and remote, allowing the user to either save the data to a local file or get a python iterator object. For local files, output files can be converted to different formats than the sliced file (CRAM -> CRAM/BAM/SAM, VCF -> VCF (.vcf or .vcf.gz)/BCF, or BCF -> VCF (.vcf or .vcf.gz)/BCF). For remote files no format conversion (CRAM -> CRAM, VCF (.vcf.gz only) -> VCF, or BCF -> BCF)

**1. Added helper functions for re-usability and to reduce the clutter in the slicing functions**
**- get_fileformat:** add more file formats so the function can return the file formats: FASTA/FARTQ/CRAM/BAM/SAM/VCF(.vcf and .vcf.gz)/BCF
**- make_pysam_inFile:** used to create a pysam AlignmentFile or VariantFile  for CRAM or VCF/BCF input files respectively. 
**- bytesio_to_iterator:** refactor to use make_pysam_inFile().
**-iter_to_file:** write an AlignmentSegment or VariantRecord iterator to local file (CRAM/BAM/SAM/VCF/BCF)

**2. Replace slice_cram and slice_remote_cram with slice_genomics and slice_remote_genomics**
refactor the slicing cram functions to handle VCF and BCF files, and renaming them to *_genomics, instead of *_cram.

**3. Replace stream_cram and save_cram class methods with stream_genomics and save_genomics**
refactor the MODO class methods stream/save_cram to handle VCF and BCF files as well, and rename them to stream/save_genomics.


modo/api.py
```diff
@@ -407,37 +415,42 @@ def stream_cram(
...
+                path=file_path,
                 region=region,
                 reference_filename=reference_filename,
             )
 
-        return cram_iter
+        return gen_iter
 
-    def save_cram(
+    def save_genomics(
```
almutlue, line 426: __Suggestion__: The fact that you had to adapt the same code twice here makes me wonder, if we really need two separate functions for saving and streaming? What about providing `modo.stream_genomics()` with an optional argument to save the output as file only?

modo/cram.py
```diff
@@ -5,86 +5,117 @@
 from pysam import (
     AlignedSegment,
     AlignmentFile,
+    VariantFile,
```
almutlue, line 8:
```suggestion
```
I think this is not used?

modo/cram.py
```diff
@@ -5,86 +5,117 @@
...
+    # path = url.path
+    url = url._replace(
+        path=re.sub("\.vcf\.\w+$", ".vcf", url.path)
+    )  # remove aditional
+    # extensions, e.g., .vcf.gz -> .vcf
     url = url._replace(path=str(Path(url.path).with_suffix("")))
 
     if region:
         reference_name, start, end = parse_region(region)
     else:
```
almutlue, line 84: __Suggestion__: As this ifelse statement occurs multiple times (see `slice_genomics()`) it could be moved into `parse_region()`.

modo/api.py
AssafSternberg: They were split originally because the behavior is different - in one case saving to a file with a None return, and in the other case returning an iterator. We can combine them with an optional output_file argument (as is the case with the local/remote slice_genomics functions, though in that case these are more for us and not the user, so less chance of the user misusing them) but there is a higher chance of the user getting confused.

modo/cram.py
```diff
@@ -5,86 +5,116 @@
...
-from .helpers import parse_region, bytesio_to_alignment_segments
-from io import BytesIO
+from .helpers import (
+    parse_region,
+    get_fileformat,
+    make_pysam_inFile,
+    bytesio_to_iterator,
+    iter_to_file,
+)
+from io import BytesIO, StringIO
```
AssafSternberg, line 23:
```suggestion
from io import BytesIO
```

remove unused import of StringIO

modo/helpers.py
```diff
@@ -226,62 +226,61 @@ def parse_region(region: str) -> tuple[str, Optional[int], Optional[int]]:
...
-        ) as alignment_file:
-            # Iterate over the alignments in the file
-            for alignment in alignment_file:
-                yield alignment
+def make_pysam_inFile(
+    path: str, fileformat: str, reference_filename: Optional[str] = None
+):
+    """Create a pysam AlignmentFile of VariantFile"""
+    if fileformat == "CRAM":
+        pysamFile = AlignmentFile(
```
almutlue, line 265: __nitpick__: Same as above - can we stick to snake case `pysamFile` --> `pysam_file`?

modo/cram.py
AssafSternberg: Good point. 

Done - see commit https://github.com/sdsc-ordes/modo-api/commit/0702e34383980e4f31b701f7f73b6052dca36d3b.

modo/helpers.py
```diff
@@ -226,62 +226,61 @@ def parse_region(region: str) -> tuple[str, Optional[int], Optional[int]]:
...
-        temp_file.seek(0)
-
-        # Open the temporary file as a pysam.AlignmentFile object
-        with AlignmentFile(
-            temp_file.name, "rc", reference_filename=reference_filename
-        ) as alignment_file:
-            # Iterate over the alignments in the file
-            for alignment in alignment_file:
-                yield alignment
+def make_pysam_inFile(
```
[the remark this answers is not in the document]
AssafSternberg: Makes sense.

Done - commit https://github.com/sdsc-ordes/modo-api/commit/3b1ef43cd78d4e032d6d75cd4307b0e8a80b2d8a

modo/io.py
almutlue, line 1: __todo__: `slice_cram` line 107 needs to be adapted to `slice_genomics`. Do we need the `slice function at all any longer?

almutlue (commented):
Thanks a lot @AssafSternberg. A nice PR and I only had minor comments. 
`modo-schema` also needs to be updated via poetry for vcfs to be accepted as data format. 
Otherwise looks fine to me!

modo/io.py
AssafSternberg: Removed, see commit https://github.com/sdsc-ordes/modo-api/commit/86aa4b2faed59601fd99a9434796f59ee5cc520b

modo/helpers.py
```diff
@@ -291,14 +301,48 @@ def bytesio_to_iterator(
...
+        )
+    else:
+        raise ValueError(
+            "Unsupported output file type. Supported files: .cram, .bam, .sam, .vcf, .vcf.gz, .bcf."
+        )
+
+    for read in gen_iter:
+        output.write(read)
+    output.close()
+    return None
```
cmdoret, line 348:
```suggestion
```
**suggestion:** if this function never returns anything, we may as well drop the return statement altogether

modo/api.py
```diff
@@ -378,79 +379,51 @@ def enrich_metadata(self):
...
 
-    def stream_cram(
+    def stream_genomics(
         self,
-        cram_path: str,
+        file_path: str,
         region: Optional[str] = None,
         reference_filename: Optional[str] = None,
+        output_filename: Optional[str] = None,
     ):
```
cmdoret, line 388: **suggestion:** would be good to add the return type hint here

cmdoret: **comment:** I think the type would be  ) `Optional[Iterator[AlignedSegment | VariantRecord]]`

modo/helpers.py
AssafSternberg: Done, see https://github.com/sdsc-ordes/modo-api/commit/6633176f4e69390aa7afe4cd925b791a1a94bc9a

modo/api.py
AssafSternberg: Done, see https://github.com/sdsc-ordes/modo-api/commit/6633176f4e69390aa7afe4cd925b791a1a94bc9a

modo/helpers.py
AssafSternberg: Done both.

Remove return None - see https://github.com/sdsc-ordes/modo-api/commit/2f5ee25f4dce980886b32f5740d91cb9a2cbfd99

Update modo-schema via poetry - see https://github.com/sdsc-ordes/modo-api/commit/d1ae6d8bb7438a35d1aebcd8f07a803ae511a03f

modo/api.py
AssafSternberg: Added - see https://github.com/sdsc-ordes/modo-api/commit/2f5ee25f4dce980886b32f5740d91cb9a2cbfd99, https://github.com/sdsc-ordes/modo-api/commit/cb3733b276b53600dbd2f51fed4d49cc36eb2e2a, and https://github.com/sdsc-ordes/modo-api/commit/2576f11d859baebb1a460b2f9efe5481891d4bd2

almutlue (approved):
Thanks a lot. Looks very good

PR #52, opened 2024-05-29
fix: mutable args

### LINKED ISSUE #47: fix: typo in the docs
Fix typo in "Find a specific MODO ad get it’s S3 path" 'ad' --> 'and'

### LINKED ISSUE #50: remove mutable default values from function arguments
We use mutable dictionaries as default values in `modo.api.MODO.__init__` and `modo.io.build_modo_from_file` e.g. :

https://github.com/sdsc-ordes/modo-api/blob/65dbc0ee563c133e2ea514bb5e02821323ea1333/modo/api.py#L55

Altering the dictionary in the function body affects the next calls, resulting in undefined behaviour.

Related bug: Instantiating `MODO` twice with `s3_endpoint` raises `TypeError`.

```python
In [1]: from modo.api import MODO
In [2]: m1 = MODO('modo-demo/ex', s3_endpoint='http://localhost/s3')

In [3]: m2 = MODO('modo-demo/ex', s3_endpoint='http://localhost/s3')
--------------------------------------------------------------------------
TypeError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 m2 = MODO('modo-demo/ex', s3_endpoint='http://localhost/s3')

File ~/Repos/github.com/sdsc-ordes/modo-api/modo/api.py:74, in MODO.__init
__(self, path, s3_endpoint, s3_kwargs, htsget_endpoint, id, name, descript
ion, creation_date, last_update_date, has_assay, source_uri)
     72 self.path = Path(path)
     73 if s3_endpoint:
---> 74     fs = s3fs.S3FileSystem(endpoint_url=s3_endpoint, **s3_kwargs)
     75     if fs.exists(str(self.path / "data.zarr")):
     76         s3_kwargs["endpoint_url"] = s3_endpoint

TypeError: s3fs.core.S3FileSystem() got multiple values for keyword argume
nt 'endpoint_url'
```

cmdoret:
This fixes #50 by using `None` as default value for function arguments instead of mutable values. [This should avoid potential bugs](https://github.com/satwikkansal/wtfpython?tab=readme-ov-file#-beware-of-default-mutable-arguments). 

In addition, it adds a test that initializes multiple remote MODOs (failed before).

edit: also fixed #47


modo/api.py
```diff
@@ -71,12 +71,13 @@ def __init__(
         self.htsget_endpoint = htsget_endpoint
         self.path = Path(path)
         if s3_endpoint:
-            fs = s3fs.S3FileSystem(endpoint_url=s3_endpoint, **s3_kwargs)
+            s3_opts = s3_kwargs or {"anon": True}
+            fs = s3fs.S3FileSystem(endpoint_url=s3_endpoint, **s3_opts)
             if fs.exists(str(self.path / "data.zarr")):
-                s3_kwargs["endpoint_url"] = s3_endpoint
+                zarr_s3_opts = s3_opts | {"endpoint_url": s3_endpoint}
```
almutlue, line 77: __Praise__: Good point and clarification! Also elegant way to concatenate options :+1:

tests/conftest.py
```diff
@@ -91,13 +91,15 @@ def remove_container():
     request.addfinalizer(remove_container)
     client = minio.get_client()
     client.make_bucket("test")
+    yield {"minio": minio}
```
almutlue, line 94: __Question(non-blocking)__: What is the advantage/reason of the `yield` here? Is it functionally required or for code clarity?

tests/test_api_remote.py
```diff
@@ -7,6 +7,21 @@
 import modo_schema.datamodel as model
 import pytest
 
+## Instantiate multiple MODOs
+
+
+@pytest.mark.slow
+def test_multi_modos(setup):
```
almutlue, line 14: __Praise__: Great to add a test for this fix as well!

almutlue (approved):
Thanks a lot - very clean and nice PR. Works smoothly on my side as well and fixes the issue.

almutlue:
I hope it was fine to add https://github.com/sdsc-ordes/modo-api/pull/52/commits/0dd9149ab879530035c1db94d00d02ec576829ad. It seemed to minor for a separate PR.

tests/conftest.py
cmdoret: In this case it's the equivalent, but in general using yield in fixtures is useful because it allows putting "cleanup code" after the yield.

When the function calling the fixture closes, `yield` gives control back to the fixture, whereas return does not.

Normally, yield gives a generator back, but I think (?) that pytest does some magic for fixtures to unwrap the value.

https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#yield-fixtures-recommended

cmdoret:
> I hope it was fine to add https://github.com/sdsc-ordes/modo-api/pull/52/commits/0dd9149ab879530035c1db94d00d02ec576829ad. It seemed to minor for a separate PR.

Thanks, I did not see it!

PR #62, opened 2024-06-07
fix(server): reload minio filesystem for each server query

### LINKED ISSUE #59: /meta endpoint broken in REST API
After renaming the modos package it seems that /meta returns an empty list when queried.

Requirements:
* [x] reproduce bug
* [x] fix

cmdoret:
The modo server did not register MODOs created after startup. This was due to the s3Fs object caching the filesystem state on instantiation by default.

This adds the `refresh=True` option to always refresh the filesystem when the server receives http requests.

Fixes #59


almutlue (approved):
Very nice and well spotted! Also works on my side with a local minio deployment and `modos.remote.list_remote_items("http://localhost")`

PR #67, opened 2024-06-12
np: Review comments from `gabriel.nuetzi`

gabyx:
# Description

These are some review comments I just came across when looking through the project with @almutlue  we could discuss.


cmdoret:
Recommendations documented in https://github.com/sdsc-ordes/modos-api/issues/122

PR #68, opened 2024-06-14
Refactor/remote

almutlue:
## Description:
Use a new `Storage` class to handle different storage protocols (in this case S3, local filesystem) within `MODO`. This simplifies the modo class code and provides a clean interface in case we need to add other storage protocols. 
`MODO` can make use of the `Storage` instance via composition.

code: @cmdoret  and @almutlue

### DIFF
--- a/modos/api.py
+++ b/modos/api.py
@@ -78,7 +78,7 @@ class MODO:
         else:
             self.storage = LocalStorage(path)
         # Opening existing object
-        if not "id" in self.zarr.attrs.keys():
+        if self.storage.empty():
             self.id = id or self.path.name
             fields = {
                 "@type": "MODO",
--- a/modos/api.py
+++ b/modos/api.py
@@ -357,7 +357,6 @@ class MODO:
         if Path(file_path) not in self.list_files():
             raise ValueError(f"{file_path} not found in {self.path}.")
 
-
         if self.htsget_endpoint:
             match GenomicFileFormat.from_filepath(Path(file_path)).name:
                 case "CRAM":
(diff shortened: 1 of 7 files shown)


modos/api.py
```diff
@@ -65,29 +68,17 @@ def __init__(
...
         else:
-            fs = None
+            self.storage = LocalStorage(path)
         # Opening existing object
-        if (self.path / "data.zarr").exists():
-            self.archive = zarr.convenience.open(str(self.path / "data.zarr"))
-        # Creating from scratch
-        else:
-            self.archive = init_zarr(self.path, fs)
+        if not "id" in self.zarr.attrs.keys():
```
almutlue, line 81: __Question__: I replaced the `storage.exists()` method here compared to our first draft, because `ZARR_ROOT` will always exists after initializing the storage. Maybe you have a better/cleaner solution than checking for the attribute keys here?

cmdoret: Ah right! then that makes the [current] Storage.exists method useless...

Perhaps we could do exactly that in `storage.exists()` and call it here? or even just `return self.zarr.attrs.keys()`

cmdoret: Alternativerly, the method could be `storage.empty()` and return `not self.zarr.attrs.keys()`. This may be semantically more accurate.

cmdoret: added `Storage.empty() -> bool` directly implemented in the abstract base class:

https://github.com/sdsc-ordes/modos-api/blob/refactor/remote/modos/storage.py#L43-L44

It is called in `MODO.__init__()` to be more explicit (see https://github.com/sdsc-ordes/modos-api/pull/68/commits/e7c3ffe63d474db40b94ed5f8c9336dd36c1c4bb)

cmdoret (approved):
Much cleaner, I made a few changes to:
1. use `match` in enums for simpler syntax
2. compatibility with the auto index upload
3. drop unused code

PR #70, opened 2024-06-25
fix: consistent ids

almutlue:
## Aim:
Use consistent ids across modos-api. 

## Changes:
- use "full_id" `/element_type/element_id` when returning any element id and within any relationship declaration (`modos.metadata()`, `has_part` relationships,`part_of` relationships)
- adapt tests
- adapt default uri in `modos.knowledgegraph`

### DIFF
--- a/modos/helpers.py
+++ b/modos/helpers.py
@@ -34,6 +34,23 @@ def dict_to_instance(element: Mapping[str, Any]) -> Any:
     )
 
 
+def full_id(element_id: str) -> bool:
+    """Checks if an element_id contains the element type as prefix.
+
+    Examples
+    --------
+    >>> full_id("sample1")
+    False
+    >>> full_id("data/test")
+    True
+    >>> full_id("/assay/test_assay")
+    True
+    """
+    etypes = [elem.value + "/" for elem in ElementType]
+    extended_etypes = etypes + ["/" + etype for etype in etypes]
+    return element_id.startswith(tuple(extended_etypes))
+
+
 def set_haspart_relationship(
     child_class: str,
     child_path: str,
@@ -74,7 +91,8 @@ def update_haspart_id(
             haspart_type = get_slot_range(has_part)
             type_name = ElementType.from_model_name(haspart_type).value
             updated_ids = [
-                f"{type_name}/{id}" for id in getattr(element, has_part)
+                id if full_id(id) else f"{type_name}/{id}"
+                for id in getattr(element, has_part)
             ]
             setattr(element, has_part, updated_ids)
     return element
--- a/modos/helpers.py
+++ b/modos/helpers.py
@@ -91,7 +91,7 @@ def update_haspart_id(
             haspart_type = get_slot_range(has_part)
             type_name = ElementType.from_model_name(haspart_type).value
             updated_ids = [
-                id if full_id(id) else f"/{type_name}/{id}"
+                id if is_full_id(id) else f"/{type_name}/{id}"
                 for id in getattr(element, has_part)
             ]
             setattr(element, has_part, updated_ids)
--- a/modos/helpers.py
+++ b/modos/helpers.py
@@ -91,7 +91,7 @@ def update_haspart_id(
             haspart_type = get_slot_range(has_part)
             type_name = ElementType.from_model_name(haspart_type).value
             updated_ids = [
-                id if full_id(id) else f"{type_name}/{id}"
+                id if full_id(id) else f"/{type_name}/{id}"
                 for id in getattr(element, has_part)
             ]
             setattr(element, has_part, updated_ids)
--- a/modos/helpers.py
+++ b/modos/helpers.py
@@ -91,7 +91,7 @@ def update_haspart_id(
             haspart_type = get_slot_range(has_part)
             type_name = ElementType.from_model_name(haspart_type).value
             updated_ids = [
-                id if is_full_id(id) else f"/{type_name}/{id}"
+                id if is_full_id(id) else f"{type_name}/{id}"
                 for id in getattr(element, has_part)
             ]
             setattr(element, has_part, updated_ids)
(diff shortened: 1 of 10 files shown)


modos/helpers.py
```diff
@@ -74,7 +91,8 @@ def update_haspart_id(
             haspart_type = get_slot_range(has_part)
             type_name = ElementType.from_model_name(haspart_type).value
             updated_ids = [
-                f"{type_name}/{id}" for id in getattr(element, has_part)
+                id if full_id(id) else f"/{type_name}/{id}"
```
cmdoret, line 94:
```suggestion
                id if is_full_id(id) else f"/{type_name}/{id}"
```

**nitpick:** reflect output in function name

modos/helpers.py
```diff
@@ -34,6 +34,23 @@ def dict_to_instance(element: Mapping[str, Any]) -> Any:
...
+    >>> full_id("sample1")
+    False
+    >>> full_id("data/test")
+    True
+    >>> full_id("/assay/test_assay")
+    True
+    """
+    etypes = [elem.value + "/" for elem in ElementType]
+    extended_etypes = etypes + ["/" + etype for etype in etypes]
+    return element_id.startswith(tuple(extended_etypes))
```
cmdoret, line 51:
```suggestion
def is_full_id(element_id: str) -> bool:
    """Checks if an element_id contains the element type as prefix.

    Examples
    --------
    >>> is_full_id("sample1")
    False
    >>> is_full_id("data/test")
    True
    >>> is_full_id("/assay/test_assay")
    True
    """
    etypes = [elem.value + "/" for elem in ElementType]
    extended_etypes = etypes + ["/" + etype for etype in etypes]
    return element_id.startswith(tuple(extended_etypes))
```
**nitpick:** reflect output in function name

cmdoret (changes requested):
The changes look good overall, but could it be that there is an issue with the `/` changes? When running `modos publish data/ex`, URI objects in the graph are missing a slash:

```ttl
<file://ex/data/demo1> a modos:DataEntity ;
    schema1:description "Dummy data for tests" ;
    schema1:name "Demo 1" ;
    modos:data_format EDAM:format_3462 ;
    modos:data_path "file://exdemo1.cram" ;
    modos:has_reference <file://exreference1> ;
    modos:has_sample <file://exsample1> .
```

almutlue:
> The changes look good overall, but could it be that there is an issue with the `/` changes?
> 
> When running `modos publish data/ex`, URI objects in the graph are missing a slash:
> 
> ```turtle
> <file://ex/data/demo1> a modos:DataEntity ;
>     schema1:description "Dummy data for tests" ;
>     schema1:name "Demo 1" ;
>     modos:data_format EDAM:format_3462 ;
>     modos:data_path "file://exdemo1.cram" ;
>     modos:has_reference <file://exreference1> ;
>     modos:has_sample <file://exsample1> .
> [...]
> ```
> 
> Note that this also happens when explicitely specifying a base URI:
> 
> ```shell
> $ modos publish --base-uri http://example.org/ data/ex
> ```
> 
> ```turtle
> <http://example.org//assay/assay1> a modos:Assay ;
>     schema1:description "Dummy assay for tests." ;
>     schema1:name "Assay 1" ;
>     modos:has_data <http://example.org/data/calls1>,
>         <http://example.org/demo1> ;
>     modos:has_sample <http://example.org/sample1> ;
>     modos:omics_type NCIT:C84343 .
> [...]
> ```
> 
> I However, I think the issue here is that there is an **extraneous** slash in the subject (not missing in the object):
> 
>     * afaik, a prefix should always end in # or /
> 
>     * when expanding with base-uri, we just do prefix+id

That is a very good point @cmdoret! I think this is a mix of problems:
1. If the prefix ends with / and our long ids start with / (e.g. /sample/sample1), prefix+id will have // as in: 
```
http://example.org//assay/assay1> a modos:Assay ;
```
Because of this I had removed the `/` from the default prefix in `modo.knowledgegraph`. But as you pointed out this causes inconsistencies with the modo itself, which has an id without starting `/` and is bad practice for the prefix itself.
So ids without a starting `/` would be more coherent with `modos publish`/`modo.knowledge_graph()`? I could update everything to `long_ids = element_type/short_id`? What do you think? 

2. I think the inconsistencies with objects and subjects in `modos publish data/ex` are partly caused by not updated `has_part` attributes in `data/ex`. I will update `data/ex` once we decided for long ids with or without starting `/`

cmdoret:
Thanks for clarifying! Based on these findings, I agree that ids should _not_ start with `/`: the separator between prefix and ID is normally defined as part of the prefix.

almutlue:
I hope this should be fine now! Last changes:
- everything should use long ids without starting `/`, so `long_id=element_type/short_id`.
- `data/ex` attributes were updated to the new long ids

`modos publish` should work as expected now:
```
modos publish data/ex
@prefix EDAM: <http://edamontology.org/> .
@prefix NCIT: <http://purl.obolibrary.org/obo/NCIT_> .
@prefix modos: <https://w3id.org/sdsc-ordes/modos-schema/> .
@prefix schema1: <http://schema.org/> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .

<file://ex/ex> a modos:MODO ;
    schema1:description "Dummy modo for tests." ;
    schema1:name "Example MODO" ;
    modos:creation_date "2024-02-19T00:00:00"^^xsd:dateTime ;
    modos:has_assay <file://ex/assay/assay1> ;
    modos:last_update_date "2024-02-19T00:00:00"^^xsd:dateTime ;
    modos:source_uri "http://example.org/ex"^^xsd:anyURI .

<file://ex/assay/assay1> a modos:Assay ;
    schema1:description "Dummy assay for tests." ;
    schema1:name "Assay 1" ;
    modos:has_data <file://ex/data/calls1>,
        <file://ex/data/demo1> ;
    modos:has_sample <file://ex/sample/sample1> ;
    modos:omics_type NCIT:C84343 .
...
``

cmdoret (approved):
Looks good and everything works as expected on my end :raised_hands: nice!

PR #71, opened 2024-06-26
fix: client-side region-filter on htsget streams

### LINKED ISSUE #51: region-based filtering not working in htsget
htsget streaming seems to return the whole region. htsget-rs receives the region specification but seems to always return all records. Reported upstream: https://github.com/umccr/htsget-rs/issues/248

cmdoret:
This PR ensures records streamed from htsget match the query region by applying lazy filter on client side. This is because the htsget server only selects record at the bgzf block level, resulting in out-of-query records.

This is a quickfix and the streaming code will need refactor as it is not maintainable nor efficient.

Fixes #51

### DIFF
--- a/modos/api.py
+++ b/modos/api.py
@@ -377,6 +377,8 @@ class MODO:
                     endpoint_type = "/reads/"
                 case "VCF" | "BCF":
                     endpoint_type = "/variants/"
+                case _:
+                    raise ValueError("Invalid file format.")
 
             # http://domain/s3 + bucket/modo/file.cram --> http://domain/htsget/reads/modo/file.cram
             # or               + bucket/modo/file.vcf.gz --> http://domain/htsget/variants/modo/file.vcf.gz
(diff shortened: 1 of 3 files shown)


almutlue (approved):
Thanks a lot! Looks good to me!

almutlue (changes requested):
Thanks a lot! Looks very good and pragmatically solves the issue!
A corner case that is not yet covered is when the region to filter for only specifies the chromosome, but not both of start and end. For example:
```
chrom, start, end = parse_region("chr22")
```
returns `chrom = chr22`, `start = None` and `end=None`.
In `bytesio_to_iterator` this causes a TypeError:
```
(get_start(record) < start) or (get_start(record) > end)
TypeError: '<' not supported between instances of 'int' and 'NoneType'
```
This could be solved with some more conditionals, but will become very nested because either `start` or `end` or `both` could be `NoneType`. Do you have a better idea?

cmdoret:
Nice catch, I fixed it with a simple solution, but it trades off performance for better readability as some unnecessary conditions are evaluated. (this is probably not the bottleneck, currently :p )

almutlue (approved):
Looks very good to me - I would also definitely prioritize readability here :smile:

PR #74, opened 2024-06-27
fix: format all relevant markdown files

gabyx:
- One could also have it in `docs` folder but needs more adjustements.


cmdoret (approved):
lgtm and readme's render properly, thanks! :+1:

PR #75, opened 2024-06-28
fea: allow virtual-host-style s3 buckets

cmdoret:
**Context:** depending on the S3 provider, buckets may be specified in path-style (domain/bucket/key), or virtual-host-style (bucket.domain/key). Modos uses botocore, which should autodetect which format is supported by the S3 server, however this detection fails depending on the provider.

**Changes:** Keep autodetect by default, but allow setting an environment variable (S3_ADDRESSING_STYLE) to force a specific bucket addressing style to "path" or "virtual".

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ WORKDIR /app
 ##################################################
 # Poetry setup
 ##################################################
-FROM python as poetry
+FROM python AS poetry
 
 WORKDIR /app
 # Install poetry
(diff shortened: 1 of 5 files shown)


almutlue (approved):
Thanks a lot!! :heart_eyes:  A very nice solution for a very tricky issue (at least from my perspective). I agree with supporting virtual host style for dedicated servers only and not further try to get both into the reverse proxy. On my side, this works smoothly (thanks for the nice instructions).

When testing, modos-server did not work properly, but returned an `ImportError`: 
```
modos-server-1  | ImportError: cannot import name 'connect_s3' from 'modos.storage' (/app/modos/storage.py)
```
This should be solved with `ghcr.io/sdsc-ordes/modos-api:latest` being updated, once this PR is merged into main

cmdoret:
Indeed, I had to build the client/server images locally for this to work, forgot to specify, sorry!

PR #76, opened 2024-07-02
feat(cli, api): remove modo

almutlue:
### Changes:
- add an api function `modo.remove_object()` to remove the entire modo object (local or remote)
- add a `--force` option to remove a modo object via cli `modos remove --force OBJECT_DIR OBJECT_NAME` as described in https://github.com/sdsc-ordes/modos-api/issues/57
- add tests for local and remote removal

### Considerations:
The api and cli logic differ - the api uses separate functions (`remove_object()` and `remove_element()`), while the cli uses the same function as for removing elements (`modos remove`). To me, separating the logic in the api seems cleaner and more readable, but I'm happy about feedback.

### DIFF
--- a/modos/api.py
+++ b/modos/api.py
@@ -25,9 +25,10 @@ from .helpers import (
     class_from_name,
     dict_to_instance,
     ElementType,
+    GenomicFileSuffix,
     set_haspart_relationship,
     UserElementType,
-    GenomicFileSuffix,
+    update_haspart_id,
 )
 from .cram import slice_genomics, slice_remote_genomics
 
@@ -45,7 +46,7 @@ class MODO:
 
     # List identifiers of samples in the archive
     >>> demo.list_samples()
-    ['/sample/sample1']
+    ['sample/sample1']
 
     # List files in the archive
     >>> files = sorted(demo.list_files())
@@ -120,7 +121,7 @@ class MODO:
         for subgroup in root.groups():
             group_type = subgroup[0]
             for name, value in list_zarr_items(subgroup[1]):
-                group_attrs[f"/{group_type}/{name}"] = dict(value.attrs)
+                group_attrs[f"{group_type}/{name}"] = dict(value.attrs)
         return group_attrs
 
     def knowledge_graph(
@@ -250,11 +251,15 @@ class MODO:
         type_group = self.zarr[type_name]
         element_path = f"{type_name}/{element.id}"
 
+        # Update part_of (parent) relationship
         if part_of is not None:
             partof_group = self.zarr[part_of]
             set_haspart_relationship(
                 element.__class__.__name__, element_path, partof_group
             )
+
+        # Update haspart relationship
+        element = update_haspart_id(element)
 
         # Add element to metadata
         attrs = json.loads(json_dumper.dumps(element))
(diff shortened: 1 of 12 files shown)


modos/api.py
```diff
@@ -196,6 +197,16 @@ def remove_element(self, element_id: str):
 
         zarr.consolidate_metadata(self.zarr.store)
 
+    def remove_object(self):
+        """Remove the complete modo object"""
+        for fi in self.list_files():
+            self.storage.remove(fi)
+        self.zarr.store.rmdir()
```
cmdoret, line 204: **praise:** very clean and satisfying, good that we abstracted this away :)

cmdoret: **comment:** also clean to do the deletion with list_files() directly. No risk of deleting any external files (when that will be possible).

modos/cli.py
```diff
@@ -188,19 +188,35 @@ def remove(
...
+                "Cannot delete root object. If you want to delete the entire MODOS, use --force."
+            )
+    else:
+        element = modo.zarr.get(element_id)
+        rm_path = element.attrs.get("data_path", [])
+        if isinstance(element, zarr.hierarchy.Group) and len(rm_path) > 0:
+            delete = typer.confirm(
+                f"Removing {element_id} will permanently delete {rm_path}.\n Please confirm that you want to continue?"
+            )
+            if not delete:
```
cmdoret, line 216: **comment:** currently, it is not possible to non-interactively delete elements within a modo.

**suggestion:** allow non-interactive deletion when `--force` is enabled. (this would require updating the parameter description).

modos/storage.py
```diff
@@ -79,9 +79,8 @@ def list(self, target: Optional[Path] = None):
                     yield file
 
     def remove(self, target: Path):
-        path = self.path / target
-        if path.exists():
-            path.unlink()
+        if target.exists():
+            target.unlink()
```
cmdoret, line 83: **question:** iiuc, this deletes by relative path instead of absolute, is this safe? Any particular reason to change it in this method specifically?

tests/test_api_remote.py
```diff
@@ -49,6 +49,28 @@ def test_remove_element(sample, remote_modo):
     assert "sample/test_sample" not in remote_modo.list_samples()
 
 
+## Remove modo
+
+
+@pytest.mark.slow
+def test_remove_modo(setup):
```
cmdoret, line 56: :smiley:   :+1:

tests/test_cli_local.py
```diff
@@ -101,3 +100,25 @@ def test_remove_element_link_list(test_modo, tmp_path):
...
+
+
+def test_not_remove_modo_without_force(test_modo, tmp_path):
+    result = runner.invoke(cli, ["remove", str(tmp_path), test_modo.path.name])
+    assert result.exit_code == 1
+    assert isinstance(result.exception, ValueError)
+    assert (
+        "Cannot delete root object. If you want to delete the entire MODOS, use --force."
+        == str(result.exception)
+    )
```
cmdoret, line 124:
```suggestion
```

**suggestion:** only test the exception type, not the message (likely to change)

cmdoret (changes requested):
Clean and simple :) I only have a few minor suggestions

modos/storage.py
almutlue: Fair point! 
My reasoning was that `modo.list_files()` returns the absolute path already. By expecting `target` to be an absolute path, the code works conveniently for remote and local storage.  
In addition in https://github.com/sdsc-ordes/modos-api/blob/e0995de6e27fe6f649c78c82ca9e6f1f182f9d1a/modos/api.py#L186-L188 we parse the absolute path. This was not working with `storage.remove(target)` removing `path = self.path / target`, but there was a bug in `modo.remove_element()`. 
So this does fix the bug and allows using `self.list_files()`independent of local or remote storage.

almutlue: I added a test to check for file deletion in https://github.com/sdsc-ordes/modos-api/pull/76/commits/7790bd22764bb3478bc9c326fdc2a7f2156b05fc for resolving this :-)

cmdoret (approved):
Perfect :raised_hands: !

PR #78, opened 2024-07-03
feat: htsget client

cmdoret:
**Context**

The htsget streaming implementation is suboptimal, as it forces greedy download of the slice and stores it on disk. This is due to limitations of the reference htsget client not allowing lazy consumption and pythonic interfaces.

**Proposed changes:**

This PR implements a minimal htsget client and refactors the streaming code to use it instead of the GA4GH implementation.

It also simplifies the streaming logic and reorders modules to improve coherence and coupling:
* introduces a `modos.genomics` module, offloading many functions from `modos.helpers`
* remove file_utils
* remove any import of `modos.api` (public api) into other modules

**Note**

For more technical details on how the htsget server and client work, see the module docstring of `modos.genomics.htsget`.

**Limitations**

This PR makes it possible to lazily consume the htsget stream, however parsing with pysam requires dumping ths whole slice to a temporary file. This is due to lack of bytesIO support in pysam (documented in https://github.com/sdsc-ordes/modos-api/pull/78/commits/ed2f8a1023fc9ab7ab3c323bfc2db252ce351cc9).

Should pysam or another library implement parsing of in-memory CRAM/BCF bufffers, it would be trivial to adapt the code.


modos/api.py
```diff
@@ -418,39 +419,71 @@ def stream_genomics(
...
+                Path(file_path), reference_filename=reference_filename
             )
-
-        return gen_iter
+            stream = (rec for rec in pysam_file)
+
+        return stream
+
+
+def build_modo_from_file(
```
cmdoret, line 443: **question:** should we just make this a classmethod (`MODO.from_file()`)?

almutlue: yes! I think that is a good idea!

cmdoret: Done!

modos/genomics/htsget.py
```diff
@@ -0,0 +1,266 @@
...
+    endpoint = format.to_htsget_endpoint()
+    stem = path.with_suffix("")
+
+    url = f"{host}/{endpoint}/{stem}?format={format.name}"
+    if region:
+        url += f"&{region.to_htsget_query()}"
+    return url
+
+
+def parse_htsget_url(url: str) -> tuple[str, Path, Optional[Region]]:
```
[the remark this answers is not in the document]
cmdoret: Will actually keep that for a separate PR if that's ok, as we don't use parse_htsget_url anywhere and it's more like a debugging helper

modos/genomics/htsget.py
```diff
@@ -0,0 +1,266 @@
...
+    """Given a URL to an htsget resource, extract the host, path, and region."""
+    parsed = urlparse(url)
+    query = parse_qs(parsed.query)
+    format = query.get("format", ["BAM"])[0]
```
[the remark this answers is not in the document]
cmdoret: indeed, vcf is default for variants, fixed!
However, I removed the default format altogether; as we already use format to infer the endpoint, we can't use the endpoint to infer the default format. (format -> endpoint is more reliable, as it's a query parameter).

modos/api.py
```diff
@@ -418,39 +423,72 @@ def stream_genomics(
...
-            url = (
-                self.htsget_endpoint
-                + endpoint_type  # /reads/ or /variants/
-                + str(Path(*Path(file_path).parts[1:]))
+            con = HtsgetConnection(
+                self.htsget_endpoint,
+                Path(*Path(file_path).parts[1:]),
+                region=_region,
             )
+            stream = con.to_pysam()
```
almutlue, line 433:
```suggestion
            stream = con.to_pysam(reference_filename=reference_filename)
```
I think this was missed during the refactoring or is there a reason for not allowing to specify a local reference for remote files? On my side python was complaining about a missing (non-downloadable) references, when printing the segments of our example modos cram file.

modos/genomics/region.py
```diff
@@ -104,6 +106,30 @@ def from_ucsc(cls, ucsc: str):
 
         return cls(chrom, start, end)
 
+    @classmethod
+    def from_pysam(
+        cls, record: pysam.VariantRecord | pysam.AlignedSegment
+    ) -> Region:
+        match record:
+            case pysam.VariantRecord:
```
almutlue, line 114:
```suggestion
            case pysam.VariantRecord():
```
Took me a while to hunt down, but will otherwise throw a type error:
```
TypeError: Argument 'other' has incorrect type (expected pysam.libcalignedsegment.AlignedSegment, got type)
``

Comment (almutlue) [inline modos/genomics/region.py:114]:
```suggestion
            case pysam.VariantRecord():
```
Took me a while to hunt down, but this will throw a type error:
```
TypeError: Argument 'other' has incorrect type (expected pysam.libcalignedsegment.AlignedSegment, got type)
```

modos/genomics/region.py
```diff
@@ -104,6 +106,30 @@ def from_ucsc(cls, ucsc: str):
...
+                chrom = record.chrom
+                start = record.start
+                end = record.stop
+            case pysam.AlignedSegment:
```
almutlue, line 118:
```suggestion
            case pysam.AlignedSegment():
```

almutlue (changes requested):
Thanks a lot @cmdoret ! Looks really good and this PR in general really is impressive. I learned about a whole new level of complexity from it!
When testing on my side, I found a very minor and hidden bug + I think an argument that you forgot to extra specify.

modos/api.py
cmdoret: Oops you're right!! thanks :)

modos/genomics/region.py
cmdoret: Ah yes, sorry I missed this one still need to get used to python's match syntax

almutlue (approved):
Perfect :rocket:

PR #82, opened 2024-07-17
chore(cli,api): rename data path argument, change yaml structure

almutlue:
## Proposed changes:
- The `--data_file`/`data_file` argument in the `modo add`/`modo.add_element()` has been renamed to `--source_file`/`source_file` to prevent ambiguity with the `data_path` attribute in the metadata schema. 

- The yaml syntax for `modo.from_file()` has been changed to accept `metadata` and `args` for each element, as suggested in https://github.com/sdsc-ordes/modos-api/issues/58. 

- The change in the yaml structure allowed refactoring `modo.from_file()` and simplifying its logic

- `modo.from_file()` will now update elements, if they already exist. This allows to use a modified version of the same yaml file, if the user wants to adapt a `modo` or to rerun `modo.from_file()`. I think this is partly related to https://github.com/sdsc-ordes/modos-api/issues/60, but still a useful feature as it allows the user to rerun their code without getting an error, specifically when `modo.from_file()` failed due to some missing metadata attribute in the yaml file.

## Note:
A nice side effect is, that this implementation of `modo.from_file` does not build the `modo` metadata schema model, but uses `modos.api.MODO()` directly. That means that the user can profit from our set defaults, e.g. for `creation_date()` and will fix the limitations in https://github.com/sdsc-ordes/modos-api/pull/73. 

## Limitations
- `modo.update_element()` does not accept a `source_file` argument. So we don't parse any `args:` from the yaml file, when updating. That means we don't update data files, but only metadata. To do so, the user still needs to remove the element (or entire modo) and add a new element. I'm not sure, if this is good enough for now or if we should refactor `update_element()` to also allow updating linked data files?
- When testing this PR I found a bug in `modo.update_element()`. It is not updating the metadata and also does not raise an error. I opened an issue here https://github.com/sdsc-ordes/modos-api/issues/81 and will address this in a separate PR.


modos/api.py
```diff
@@ -451,44 +451,40 @@ def from_file(
...
-        if len(modo_inst) != 1:
+        element_list = parse_attributes(Path(path))
+        modo_count = len(
+            [
+                ele
+                for ele in element_list
+                if ele["metadata"].get("@type") == "MODO"
+            ]
+        )
+        if modo_count != 1:
```
almutlue, line 462: I don't really like this code, but I thought it could be a useful check that should not be removed, just to keep the code cleaner. Happy about any suggestion?

cmdoret: I agree it's good to check it, and the code is fine IMO. A slightly shorter variant would be this, but this is basically the same.

```python
sum(
    [
        ele["metadata"].get("@type") == "MODO"
        for ele in element_list
    ]
)
```

or in functional style:
```python
sum(
    map(
        lambda e: e['metadata'].get('@type') == 'MODO',
        element_list
    )
)
```

modos/api.py
```diff
@@ -451,44 +451,40 @@ def from_file(
         htsget_endpoint: Optional[str] = None,
     ) -> MODO:
         """build a modo from a yaml or json file"""
-        instances = parse_multiple_instances(Path(path))
-        # check for unique ids and fail early
-        ids = [inst.id for inst in instances]
-        if len(ids) > len(set(ids)):
```
cmdoret, line 457: **question:** is this check not needed anymore? Do I understand correctly that in the case of multiple elements with the same id in the file, we will silently override previous ones?

cmdoret (commented):
Nice! :+1:  I only have some high level questions related to naming and checks

data/ex_config.yaml
```diff
@@ -1,35 +1,52 @@
...
+    last_update_date: "2024-01-17T00:00:00"
+    has_assay: assay1
 
-- id: assay1
-  "@type": Assay
-  name: Assay 1
-  description: Example assay for tests
-  has_sample: sample1
-  omics_type: GENOMICS
+- metadata:
```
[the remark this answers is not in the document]
almutlue: Good point! I changed the key in yaml and the cli argument to `element` as we use metadata in several contexts throughout the code already.

data/ex_config.yaml
```diff
@@ -1,35 +1,52 @@
...
-  data_format: CRAM
-  data_path: data/ex/demo1.cram
-  has_reference: reference1
+- metadata:
+    id: demo1
+    "@type": DataEntity
+    name: Demo 1
+    description: Demo CRAM file for tests.
+    data_format: CRAM
+    data_path: demo1.cram
```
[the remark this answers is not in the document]
almutlue: It is a nice idea and only a few lines of additional code (at least for `MODO.from_file()`).
For the `cli` it is more tedious, because `data_path` belongs to the required slots when prompting. I would suggest that there we keep it as it is for now?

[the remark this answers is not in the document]
cmdoret: makes sense, tracked in issues https://github.com/sdsc-ordes/modos-api/issues/85

docs/tutorials/modo_provider.md
```diff
@@ -135,15 +135,15 @@ Alternatively, a MODO and all associated elements can be specified in a `yaml-fi
...
+- element:
     id: sample1
     "@type": Sample
     name: Sample 1
     description: An example sample for tests.
     collector: Foo university
     sex: Male
 ```
-In this yaml-file each element is a separate list entry. Within each list entry the `metadata` section specifies all relevant metadata as well as the `element type` (using `@type: ELEMENT_TYPE` as syntax).
+In this yaml-file each element is a separate list entry. Within each list entry the `element` section specifies all relevant metadata as well as the `element type` (using `@type: ELEMENT_TYPE` as syntax).
```
cmdoret, line 146:
```suggestion
In this yaml file each element is a separate list entry. Within each list entry the `element` section specifies all relevant metadata as well as the `element type` (using `@type: ELEMENT_TYPE` as syntax).
```

cmdoret (approved):
It looks good like this! :+1:

PR #87, opened 2024-07-24
feat(cli): stream command

cmdoret:
Adds a stream command that sends the binary data from a remote HTS file into stdout. This enables piping modos genomic files into other tools.

Questions:
* Currently only works with remote files. Does it make sense to make it work with local files as well?
* Currently takes eleemnt_id as input. Would it make more sense to take a file path or not?

### DIFF
--- a/modos/cli.py
+++ b/modos/cli.py
@@ -360,12 +360,11 @@ def publish(
 
 @cli.command()
 def stream(
-    object_directory: Annotated[Path, typer.Argument(...)],
-    element_id: Annotated[
+    file_path: Annotated[
         str,
         typer.Argument(
             ...,
-            help="The identifier within the modo. Use modos show to check it.",
+            help="The path to the file to stream . Use modos show --files to check it.",
         ),
     ],
     s3_endpoint: Annotated[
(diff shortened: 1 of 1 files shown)


almutlue (commented):
Really nice and clean thanks to the client refactoring :star_struck: !
Regarding your questions: 

1. Good question, for consistency, it would be nice to work with local files as well, but the whole point of streaming and using modos for slicing does not make much sense locally. So I think both is fine. 

2. What about adding another argument `--data-path` or `--file-path` to allow the user to specify either the `element_id` or the `file-path`?

cmdoret:
> Really nice and clean thanks to the client refactoring 🤩 ! Regarding your questions:
> 
>     1. Good question, for consistency, it would be nice to work with local files as well, but the whole point of streaming and using modos for slicing does not make much sense locally. So I think both is fine.
> 
>     2. What about adding another argument `--data-path` or `--file-path` to allow the user to specify either the `element_id` or the `file-path`?

Thanks @almutlue  :) 

1. OK, will keep as-is then, less code is better.
2. We could do it in two ways:
  + `--file-path` flag changes the nature of the argument.
  + remove argument, add two mutually exclusive options.

For 2. I feel like both solutions make the CLI a bit confusing (kind-of stateful), I should we stick to one or the other until we figure out a clean solution?

almutlue:
> > Really nice and clean thanks to the client refactoring 🤩 ! Regarding your questions:
> > ```
> > 1. Good question, for consistency, it would be nice to work with local files as well, but the whole point of streaming and using modos for slicing does not make much sense locally. So I think both is fine.
> > 
> > 2. What about adding another argument `--data-path` or `--file-path` to allow the user to specify either the `element_id` or the `file-path`?
> > ```
> 
> Thanks @almutlue :)
> 
>     1. OK, will keep as-is then, less code is better.
> 
>     2. We could do it in two ways:
> 
> 
>     * `--file-path` flag changes the nature of the argument.
> 
>     * remove argument, add two mutually exclusive options.
> 
> 
> For 2. I feel like both solutions make the CLI a bit confusing (kind-of stateful), I should we stick to one or the other until we figure out a clean solution?

Fair point. I don't have a strong opinion, but maybe a slight preference for `file-path`, because in `stream_genomics()` in the api we use the file path as input argument and it is not explicit that `element_id` needs to refer to an element with a `data_path` attribute?

almutlue (approved):

cmdoret:
Sounds good to me, changed the argument to filepath and simplified the command a bit by taking the full path as a single argument:

```sh
modos stream -s3 <endpoint> bucket/modos/demo1.cram
```

PR #90, opened 2024-07-26
feat: streamline remote options

cmdoret:
**Objective:**

Allows users to always specify the same modos server endpoint. Underlying htsget/s3 endpoints are auto-detected internally by querying the modos server.

For testing / advanced usage, a `services` option is still exposed in the API to bypass this and explicitely provide service endpoints.

**Notes**

* As every command in the CLI was copy-pasting the `s3_endpoint` option, this PR removes it in favor of a top-level `endpoint` flag that is inherited by all commands (through typer.Context).
* `endpoint` option can also read from the environment variable `MODOS_ENDPOINT`.
* the docker compose config used a static network ip. This was only needed because we passed the S3_LOCAL_URL to htsget instead of S3_PUBLIC_URL (by mistake). Fixing the env var makes the static IP unnecessary, and this PR removes it.
* `pydantic.HttpUrl` was used in some places for more explicit type hints, this PR adds it throughout the package and decorates functions with `@validate_call`, to validate url params at runtime.
* expose the version from `pyproject.toml` through `modos.__version__` and in the cli through `modos --version` using a typer callback.
* describe MODO arguments in docstring

**Limitations**

* If the environment variable is set, the user has to manually override it to create local objects which may not be intuitive.
* This PR implements runtime parameter validation on functions (via `pydantic.validate_call`) and dataclases (by replacing `dataclasses.dataclass` with `pydantic.dataclasses.dataclass`), but not on normal classes.

**Questions**
should we use a more explicit mechanism for remote/local paths to avoid confusion? (e.g. `--local` flag or explicitely requiring `s3://` paths for remote objects.). Should this be in this, or a separate PR?

**Examples**

CLI:
```shell
$ modos --version
modos 0.1.0

# create remote object on the s3 provided by the modos server at http://localhost
$ modos --endpoint http://localhost create bucket/object

# read endpoint from env variable to avoid repetition
$ export MODOS_ENDPOINT='http://localhost'
$ modos create bucket/object2
$ modos show bucket/object2
$ modos delete bucket/object2
```

API:
```python
>>> from modos.api import MODO
>>> MODO('bucket/object2', endpoint='http://localhost')

# This is also possible, e.g. user has an s3 server, but no modos server
>>> nostream = MODO('bucket/ex-nostream', services={'s3': 'http://s3.example.org'})
>>> nostream.stream_genomics('demo1.cram')
ValueError: No htsget endpoint provided
```

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -1,5 +1,3 @@
-version: '3.9'
-
 volumes:
   minio-data:
     driver: local
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -6,11 +6,7 @@ volumes:
 
 networks:
   modos-network:
-    external: false
-    ipam:
-      config:
-        - subnet: 172.23.0.0/16
-          gateway: 172.23.0.1
+    driver: bridge
 
 services:
 
@@ -29,8 +25,7 @@ services:
     ports:
       - "80:80"
     networks:
-      modos-network:
-        ipv4_address: 172.23.0.5
+      - modos-network
 
   minio:
     image: docker.io/bitnami/minio:2024
@@ -41,8 +36,7 @@ services:
     volumes:
       - minio-data:/bitnami/minio/data
     networks:
-      modos-network:
-        ipv4_address: 172.23.0.2
+      - modos-network
     environment:
       - MINIO_ROOT_USER=${MINIO_ROOT_USER:-minio}
       - MINIO_ROOT_PASSWORD=${MINIO_ROOT_PASSWORD:-miniosecret}
@@ -62,10 +56,9 @@ services:
       - "8080" # ticket server
       - "8081" # data block server
     networks:
-      modos-network:
-        ipv4_address: 172.23.0.3
+      - modos-network
     environment:
-      - ENDPOINT=${S3_LOCAL_URL:-http://172.23.0.2:9000}
+      - ENDPOINT=${S3_PUBLIC_URL:-http://localhost/s3}
       - BUCKET=${S3_BUCKET:-modos-demo}
       - S3_ADDRESSING_STYLE=${S3_ADDRESSING_STYLE:-auto}
       - AWS_ACCESS_KEY_ID=${MINIO_ROOT_USER:-minio}
@@ -79,8 +72,7 @@ services:
     expose:
       - "8000"
     networks:
-      modos-network:
-        ipv4_address: 172.23.0.4
+      - modos-network
     environment:
       - S3_LOCAL_URL=${S3_LOCAL_URL:-http://minio:9000}
       - S3_PUBLIC_URL=${S3_PUBLIC_URL:-http://localhost/s3}
--- a/modos/cli.py
+++ b/modos/cli.py
@@ -7,7 +7,6 @@ from enum import Enum
 import os
 from pathlib import Path
 from typing import Any, List, Mapping, Optional
-from numpy import source
 from typing_extensions import Annotated
 
 import click
@@ -25,6 +24,7 @@ from .helpers.schema import (
     get_slot_range,
     load_schema,
 )
+from .remote import list_endpoints
 from .genomics.htsget import HtsgetConnection
 from .genomics.region import Region
 from .io import parse_instance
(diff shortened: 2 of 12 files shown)


modos/cli.py
```diff
@@ -142,24 +139,24 @@ def create(
...
-    # Initialize object's directory
-    if object_directory.exists():
-        raise ValueError(f"Directory already exists: {object_directory}")
 
-    if s3_endpoint:
-        fs = connect_s3(s3_endpoint, {"anon": True})
+    endpoint = ctx.obj.endpoint
+    # Initialize object's directory
+    if endpoint:
+        s3 = list_endpoints(endpoint)["s3"]  # type: ignore
```
almutlue, line 146: __todo__: `list_endpoints()` is not defined and `modos create --endpoint` will break.
Maybe worth implementing in `modos.remote`?
Or something like this:
```
s3 = EndpointManager(endpoint).s3
```

almutlue (changes requested):
Really nice PR with a lot of small improvements. I learned a lot from reviewing it, e.g. I didn't know about `@validate_call` and `typer.context` :star_struck: !

I just have one comment about `list_endpoints()`, which looks like you planned, but forgot to implement.

Regarding your question on using a scheme or flag to indicate local/remote usage I'm unsure and keep changing my mind. A scheme would require fewer arguments, but I like the current implementation, that does only require `MODO_ENDPOINT` to be set for working remotely and no additional specification :thinking: .

cmdoret:
TODO:
* make only use remote modos when `s3://` scheme is specified (otherwise local
* add some info logs (with print) to stderr

modos/cli.py
cmdoret: df7ff1276b01fa1bc1ed72066fd5e215730bc26c :heavy_check_mark:

deploy/docker-compose.yml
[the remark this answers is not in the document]
cmdoret: c734402 :heavy_check_mark:

cmdoret:
Added this based on our discussion:
* rudimentary logs on MODO instantiation
* require `s3://` for remote paths
  + check done on MODO instantiation using a custom S3Path pydantic model
  + changes made in `modos.storage.S3Storage` to support this

modos/cli.py
```diff
@@ -45,6 +45,14 @@ class RdfFormat(str, Enum):
 
 cli = typer.Typer(add_completion=False)
 
+OBJECT_PATH_ARG = Annotated[
+    Path,
```
almutlue, line 49: __suggestion(blocking)__:
```suggestion
    str,
```
Parsing as `PosixPath` is going to break `modos create --endpoint http://localhost create s3://modos-demo/test`, because it converts the object_path to `s3:/modos-demo/test`. I think parsing as string would work, but maybe you have a better idea to parse both url and path?

modos/cli.py
```diff
@@ -143,20 +151,18 @@ def create(
...
         fs = connect_s3(s3, {"anon": True})  # type: ignore
-        if fs.exists(object_directory):
-            raise ValueError(
-                f"Remote directory already exists: {object_directory}"
-            )
-    elif object_directory.exists():
-        raise ValueError(f"Directory already exists: {object_directory}")
+        if fs.exists(object_path):
+            raise ValueError(f"Remote directory already exists: {object_path}")
+    elif object_path.exists():
```
almutlue, line 158:
```suggestion
    elif Path(object_path).exists():
```
Only needed, if `OBJECT_PATH_ARG` is parsed as string

almutlue (changes requested):
Looks very good! Thanks a lot @cmdoret !
I ran into an error when testing `modos --endpoint http://localhost` create s3://modos-demo/test` because of `pathlib.Path` auto converting urls to paths :exploding_head: .
Easy to overlook and hard to spot. I'm not sure I like that behaviour of pathlib ^^

modos/cli.py
cmdoret: I agree, it makes sense to use strings in the CLI and do the validation only in MODO

almutlue (approved):
Great! Thanks a lot - works perfectly on my side and makes remote usage much cleaner!

PR #95, opened 2024-08-06
ci: build and publish at pypi/test-pypi

almutlue:
### Scope:
Two new github action workflows with 2 jobs each:
- `poetry-publish`: 1. build modos-api and 2. publish at pypi, triggered on published releases
- `poetry-test-publish`:  1. build modos-api and 2. publish at test-pypi, manually triggered by workflow dispatch

The publishing job uses trusted publisher management for authentication, which requires write permission. To not elevate privileges building and publishing were separated into two different ci-jobs as recommended.

### Limitations:
So far only `poetry-test-publish` was run (see [output](https://github.com/sdsc-ordes/modos-api/actions/runs/10265083314)) and tested. Both workflows are build analogous, but I think we have to wait for the next release to check `poetry-publish`?

### DIFF
--- a/.github/workflows/poetry-publish.yml
+++ b/.github/workflows/poetry-publish.yml
@@ -59,4 +59,4 @@ jobs:
         path: dist/
     - name: Publish package distributions to PyPI
     # https://github.com/pypa/gh-action-pypi-publish
-      uses: pypa/gh-action-pypi-publish@release/v1.9.0
+      uses: pypa/gh-action-pypi-publish@release/v1
--- a/.github/workflows/poetry-publish.yml
+++ b/.github/workflows/poetry-publish.yml
@@ -13,6 +13,9 @@ permissions:
   contents: read
 
 jobs:
+  run-tests:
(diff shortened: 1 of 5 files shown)


cmdoret (approved):
This looks very good as-is, so I'm giving the :heavy_check_mark: 
**question(non-blocking):** should we run tests in this workflow before build+release to avoid releasing broken code by accident?

almutlue:
Good point - I added a `run-tests` job that imports the test workflow to both workflows. 
Both the building and publishing jobs depend on the test job to make sure that it fails early enough.

Publishing fails in [the current test-publish action](https://github.com/sdsc-ordes/modos-api/actions/runs/10280739436) because the current version already exists at test-pypi.  To solve this, it is suggested to make dev releases on pypi (see https://github.com/pypi/warehouse/issues/726), but I thought it is not necessary, because the changes should not affect publishing?

cmdoret (approved):
Very clean and DRY, :+1:

PR #96, opened 2024-08-06
fix(rdf): schema.org prefix

cmdoret:
The `schema` prefix is assigned by default by rdflib to **https**://schema.org.
Unfortunately, linkml cannot use **https**://schema.org and assigns the schema namespace to **http**://schema.org

As modos inherits **http** from linkml, all conversions to rdflib resulted in using the automatically assigned `schema1` prefix.

This PR overrides the default rdflib assignment of `schema` so that turtle serialization and SPARQL queries on modos objects can use the `schema` prefix.


almutlue (approved):
Well spotted!  Looks very good to me.

PR #98, opened 2024-08-08
feat(cli, api):show only target element

cmdoret:
Quality of life feature to target a specific element when inspecting an object:

CLI:
```shell
$ modos show data/ex sample/sample1
sample/sample1:
  '@type': Sample
  cell_type: astrocyte
  description: Dummy sample for tests
  name: Sample 1
  sex: Male
  source_material: brain tissue

$ modos show --zarr data/ex assay
assay
 └── assay1
```

API:

```python
m = MODO('data/ex')
m.show_contents('sample/sample1')
m.list_arrays('data')
```

**Limitations**

Does not work with the `--files` option, as I was not sure whether it makes sense; e.g. should this list the index files (.crai, .csi, ...), should we list files directly linked to the element, or also recursively (sample1 -> data1 -> file).

### DIFF
--- a/modos/api.py
+++ b/modos/api.py
@@ -189,10 +189,10 @@ class MODO:
         """Lists files in the archive recursively (except for the zarr file)."""
         return [fi for fi in self.storage.list()]
 
-    def list_arrays(self):
+    def list_arrays(self, element_id: Optional[str] = None):
         """Lists arrays in the archive recursively."""
         root = zarr.convenience.open_consolidated(self.zarr.store)
-        return root.tree()
+        return root[element_id or "/"].tree()
 
     def query(self, query: str):
         """Use SPARQL to query the metadata graph"""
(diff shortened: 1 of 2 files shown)


almutlue (approved):
Very nice and neat feature! Works perfectly and the code is very clean!

__Question(minor, non-blocking)__: Would it be useful to also allow subsetting at any place in the zarr tree for `modos.show_contents()` ? E.g. something like:
```shell
$ modos show data/ex assay
assay/assay1:
  '@type': Assay
  description: Dummy assay for tests.
  has_data:
  - data/demo1
  - data/calls1
  has_sample:
  - sample/sample1
  name: Assay 1
  omics_type:
  - GENOMICS
assay/assay2:
...
```
I'm not sure if it is a good/useful idea and implementation would make the code more convoluted, so feel free to ignore it.

PR #106, opened 2024-10-01
feat: support terminology codes

cmdoret:
## Summary
This adds support for terminology codes instead of free text for specific metadata fields, along with autocomplete suggestion in the terminal.

Currently, the following properties/terminologies are used:
* "cell_type": "https://purl.obolibrary.org/obo/cl.owl",
* "source_material": "https://purl.obolibrary.org/obo/uberon.owl",
* "taxon_id": "https://purl.obolibrary.org/obo/ncbitaxon/subsets/taxslim.owl"

## Major changes

* Refactor CLI code: move prompt utilities to dedicated module (modos.prompt)
* Add code-matching module (modos.codes)
* Implement CodeMatcher protocol with two members (remote/local)
  + Remote is used if an endpoint is provided (completion runs on server, faster)
  + Local used as fallback (ontology downloaded and runs on client machine)
* fuzon-http service added as a service in modos server deployment
* pyfuzon added as extra dependency for local code matching

## Trying it out

To test local autocomplete in terminal:
```sh
modos create data/example
modos add data/example sample
```

To rely on the server for autocomplete:
```sh
make deploy
modos --endpoint=http://localhost create data/example
modos --endpoint=http://localhost add data/example sample
```

## Notes

Codes are recommended based similarity between user input and labels, but only the URIs are persisted in metadata.

## Follow up (separate issues):
* speed up download when using local completer
  + terminology caching (+async download in background)

## Open questions

When creating a modos from input yaml (instead of interactively) (see [`data/ex_config.yaml`](https://github.com/sdsc-ordes/modos-api/blob/main/data/ex_config.yaml)), URIs are now required for the 3 properties above.

It may be painful for users to find out what URIs to input in the yaml. Should we provide some kind of subcommand just to get the codes (basically a fuzon wrapper)?
Perhaps something along the lines of this 
```shell
# modos codes <property> <query>
modos codes cell_type "red blood cell"
```


deploy/fuzon/Dockerfile
```diff
@@ -0,0 +1,26 @@
+FROM rust:1.75.0 AS builder
+
+WORKDIR /build
+
+RUN cargo install cargo-strip
+
+RUN apt-get update && apt-get install -y git
+
+RUN git clone https://github.com/sdsc-ordes/fuzon.git --branch feat/fuzon-http .
```
cmdoret, line 9: TODO: clone release instead of branch, once PR is merged

deploy/docker-compose.yml
```diff
@@ -15,11 +15,13 @@ services:
     volumes:
       - ./nginx/default.conf.template:/etc/nginx/templates/default.conf.template:ro
     environment:
-      - S3_PUBLIC_URL=${S3_PUBLIC_URL:-http://localhost/s3/}
-      - S3_LOCAL_URL=${S3_lOCAL_URL:-http://minio:9000}
+      - FUZON_PUBLIC_URL=${FUZON_PUBLIC_URL:-http://localhost/fuzon}
       - HTSGET_PUBLIC_URL=${HTSGET_PUBLIC_URL:-http://localhost/htsget}
+      - S3_PUBLIC_URL=${S3_PUBLIC_URL:-http://localhost/s3/}
+      - FUZON_LOCAL_URL=${FUZON_LOCAL_URL:-http://fuzon:8080}
       - HTSGET_LOCAL_URL=${HTSGET_LOCAL_URL:-http://htsget:8080}
```
supermaxiste, line 22: **question(clarification)**: from what I understand both fuzon and htsget will use the same port for each service to run. Is this fine in general? Networking is by far my weakest point, so I hope the question is not too basic.

deploy/docker-compose.yml
```diff
@@ -15,11 +15,13 @@ services:
...
       - MODOS_LOCAL_URL=${MODOS_LOCAL_URL:-http://modos-server:8000}
+      - S3_LOCAL_URL=${S3_lOCAL_URL:-http://minio:9000}
```
supermaxiste, line 24: Call me a hawk
```suggestion
      - S3_LOCAL_URL=${S3_LOCAL_URL:-http://minio:9000}
```

deploy/docker-compose.yml
```diff
@@ -64,6 +66,16 @@ services:
       - RUST_LOG=debug
 
 
+  fuzon:
+    build: ./fuzon
+    volumes:
+      - ./fuzon/config.json.template:/fuzon/config.json.template:ro
```
supermaxiste, line 72: **praise**: never saw this notation for mounts where there's three colons and TIL that you can set permissions

deploy/nginx/default.conf.template
```diff
@@ -41,4 +42,9 @@ server {
   location / {
     proxy_pass ${MODOS_LOCAL_URL} ;
   }
+  # terminology code matching server
+  location /fuzon {
+    rewrite ^/fuzon/(.*) /$1 break ;
+    proxy_pass ${FUZON_LOCAL_URL} ;
```
supermaxiste, line 48: **question(clarification)**: the config template provides both a public & local setup. How is this config used exactly to make sure that we're dealing with either public or local setup? The main reason I'm asking is if it makes more sense to separate the two into different files or make it clearer what needs to be modified in which case.

modos/codes.py
```diff
@@ -0,0 +1,68 @@
...
+
+            self.matcher = TermMatcher.from_files([SLOT_TERMINOLOGIES[slot]])
+        except ImportError:
+            raise ValueError(
+                """No endpoint provided and pyfuzon not installed,
+                cannot do code matching."""
+            )
+
+    def find_codes(self, query: str) -> list[Code]:
+        return self.matcher.top(query, 50)
```
supermaxiste, line 41: **suggestion(enhancement, non-blocking)**: we're hardcoding top50 matches if I get it correctly: any chance we can parametrize it somewhere? Can be a separate issue for another PR

modos/prompt.py
```diff
@@ -0,0 +1,130 @@
...
+            target_class
+                Class to build
+            exclude
+                Mapping with the name of a slot as key  and list of invalid entries as values.
+        """
+
+        entries = {}
+        required_slots = set(get_slots(target_class, required_only=True))
+        optional_slots = (
+            set(get_slots(target_class, required_only=False)) - required_slots
```
supermaxiste, line 105: **praise**: TIL that you can use arithmetic operations on sets which are equivalent to unions/differences/joins

supermaxiste:
## Code review 

Mostly suggestions, recommendations and clarifications.
To add to my other comments: `Makefile` L42 has a typo in the `S3_PUBLIC_URL`:

https://github.com/sdsc-ordes/modos-api/blob/8d9deb2dbe7dbc5cc9a191b18fc6e30dc94779bd/Makefile#L42

deploy/docker-compose.yml
cmdoret: sure, that can be confusing; they use the same port, but on different containers:
when on localhost:
* fuzon is `http://fuzon:8080`
* htsget is `http://htsget:8080`

However, when accessing from outside the machine, everything is accessed through a reverse proxy (see [the nginx config](https://github.com/sdsc-ordes/modos-api/blob/feat/fuzon-codes/deploy/nginx/default.conf.template#L37-L48)) which redirects incoming requests to local services as follows:
* http://<server>:80/fuzon -> fuzon:8080
* http://<server>:80/htsget -> htsget:8080

deploy/docker-compose.yml
cmdoret: daaamn good catch, I wish I had a linter for this kind of stuff :thinking:

deploy/nginx/default.conf.template
cmdoret: Documented in [06ca00b](https://github.com/sdsc-ordes/modos-api/pull/106/commits/06ca00bade6c6e0a11f79f786ea93293c48651f0) @marftn  do you think this whole thing makes sense, or is this some kind of antipattern?

modos/codes.py
cmdoret: good point, this is parametrized as of ac78fa1

modos/prompt.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+
+        # Always require identifiers if possible
```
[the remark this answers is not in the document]
cmdoret: Indeed, it's always required 270aec0

I don't think we need a warning here as id is the key used by zarr to store the object; every object must be referencable

supermaxiste:
## Deployment review 

**praise**: with `make deploy` everything was smooth
**praise**: the autocomplete feature is not only fast, but works extremely smoothly 🤓 

**issue(minor, non-blocking)**: the recommendations can sometimes point to blank nodes. When blank nodes are selected, `modos` throws an error because we're not providing a proper URI. See screenshot as an example.

Screenshot 2024-10-21 at 16 28 55
[...]
Screenshot 2024-10-21 at 16 36 29

**question(minor, non-blocking)**: when working with `modos` objects, there's no `list` command and it becomes a bit confusing to work with objects that you created but forgot about. Maybe I missed a command, but as far as I can see everything assumes that we know the objects we're working with.

```shell
$ cli:~/modo-api$ modos --help
Usage: modos [OPTIONS] COMMAND [ARGS]...

  Multi-Omics Digital Objects command line interface.

Options:
  --endpoint TEXT  URL of modos server.  [env var: MODOS_ENDPOINT]
  --version        Print version of modos client
  --help           Show this message and exit.

Commands:
  add      Add elements to a modo.
  create   Create a modo interactively or from a file.
  publish  Export a modo as linked data.
  remove   Removes an element and its files from the modo.
  show     Show the contents of a modo.
  stream   Stream genomic file from a remote modo into stdout.
  update   Update a modo based on a yaml file.
```

I will pre-approve the PR since all the points I raised might be a separate PR.

supermaxiste (approved):

cmdoret:
blank nodes: good catch! Addressed upstream, as it makes no sense for fuzon to even keep these in memory (they have an undetermined URI) https://github.com/sdsc-ordes/fuzon/pull/31

cmdoret:
I've also added a `modos search-codes` command to provide an easy way to find code URIs

PR #118, opened 2025-02-07
fix(api): update data path

almutlue:
Depends on: https://github.com/sdsc-ordes/modos-schema/pull/21

### Goal:
1. Fix https://github.com/sdsc-ordes/modos-api/issues/116 to update a data_path on an existing object when read from file.
2. Check and update files, that are part of modos, if the source-file has changed 

### Changes:
- add data-file checksum attribute by calculating the `blake2B checksum` to a `DataEntity` when it is added and has a linked file. 
- `modos.update_element()` accepts `source_file` and `part_of` args (consistent `modos.add_element()`). 
-  `modos.update_element()` moves associated files, if the data_path attribute has changed
- `modos.update_element()` updates associated files, if the `data_checksum` differs 
- `modos.storage.Storage` class includes a move method
- `modos.update_element()`: index files related to genomic files are moved and updated accordingly
- test update in local and remote context

### Questions:
- Should we use `MD5-checksum` instead as we don't use it in a security-related context? `Blake2B` should be similar efficient, but it less known?
- Could/should we remove the index logic from `modos.api.update_element()`? I think that would be cleaner, but would also require some restructuring. Maybe in another PR?

### DIFF
--- a/modos/api.py
+++ b/modos/api.py
@@ -381,6 +381,8 @@ class MODO:
         element_id: str,
         new: model.DataEntity | model.Sample | model.Assay | model.MODO,
         source_file: Optional[Path] = None,
+        part_of: Optional[str] = None,
+        allowed_elements: type = ElementType,
     ):
         """Update element metadata in place by adding new values from model object.
 
@@ -409,6 +411,15 @@ class MODO:
                 if source_checksum != attr_dict.get("data_checksum"):
                     self.storage.put(source_file, new_path)
                     new["data_checksum"] = source_checksum
+
+        type_name = allowed_elements.from_object(new).value
+        element_path = f"{type_name}/{new.id}"
+
+        if part_of is not None:
+            partof_group = self.zarr[part_of]
+            set_haspart_relationship(
+                new.__class__.__name__, element_path, partof_group
+            )
 
         new = update_haspart_id(new)
         new = json.loads(json_dumper.dumps(new))
--- a/modos/api.py
+++ b/modos/api.py
@@ -32,7 +32,7 @@ from modos.helpers.schema import (
     set_haspart_relationship,
     UserElementType,
     update_haspart_id,
-    generate_data_checksum,
+    DataElement,
 )
 from modos.genomics.formats import get_index, read_pysam
 from modos.genomics.htsget import HtsgetConnection
@@ -331,23 +331,16 @@ class MODO:
 
         # Copy data file to storage
         if source_file:
-            source_path = Path(source_file)
-            target_path = Path(element._get("data_path"))
-            self.storage.put(source_path, target_path)
+            # NOTE: Keep this for compatibility until ReferenceGenomes are handeled by refget
+            if isinstance(element, model.ReferenceGenome):
+                source_path = Path(source_file)
+                target_path = Path(element._get("data_path"))
+                self.storage.put(source_path, target_path)
 
             # Add data_checksum attribute
             if isinstance(element, model.DataEntity):
-                setattr(
-                    element,
-                    "data_checksum",
-                    generate_data_checksum(source_file),
-                )
-
-            # Genomic files have an associated index file
-            source_ix = get_index(source_path)
-            target_ix = get_index(target_path)
-            if source_ix:
-                self.storage.put(source_ix, target_ix)
+                new_data = DataElement(element)
+                new_data.process_and_store(self.storage, Path(source_file))
 
         # Infer type
         type_name = allowed_elements.from_object(element).value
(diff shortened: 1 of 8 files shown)


pyproject.toml
almutlue, line 1: `botocore` is only pinned to enable tests until https://github.com/sdsc-ordes/modos-api/pull/117 is merged

poetry.lock
almutlue, line 1: Superseded by https://github.com/sdsc-ordes/modos-api/pull/117

cmdoret:
(rebased on main)

cmdoret (commented):
Awesome work! I added a few ideas/suggestions, but nothing major :)

modos/api.py
```diff
@@ -406,6 +439,9 @@ def update_element(
         attrs.update(**new_items)
         self.update_date()
 
+        if source_file:
+            pass
```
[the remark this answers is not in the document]
almutlue: Yes, you are right. This should be deleted. Thanks for noticing!

modos/storage.py
```diff
@@ -99,6 +103,9 @@ def remove(self, target: Path):
     def put(self, source: Path, target: Path):
         shutil.copy(source, self.path / target)
 
+    def move(self, source: Path, target: Path):
+        shutil.move(self.path / source, self.path / target)
```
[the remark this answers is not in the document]
almutlue: Very good point! I renamed `source` to `rel_source` in `storage.move` for clarity. `target` is always relative to `modos.path` and I kept it as is to keep it short. I also added docstrings to the abstract class method move.

modos/api.py
```diff
@@ -328,26 +329,27 @@ def _add_any_element(
...
+            if isinstance(element, model.DataEntity):
+                setattr(
+                    element,
+                    "data_checksum",
+                    generate_data_checksum(source_file),
                 )
+
+            # Genomic files have an associated index file
+            source_ix = get_index(source_path)
+            target_ix = get_index(target_path)
```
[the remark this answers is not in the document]
almutlue: Thanks a lot. Yes, a facade class is a good idea. I gave it a try :-)

modos/api.py
```diff
@@ -331,23 +331,16 @@ def _add_any_element(
...
         # Copy data file to storage
         if source_file:
-            source_path = Path(source_file)
-            target_path = Path(element._get("data_path"))
-            self.storage.put(source_path, target_path)
+            # NOTE: Keep this for compatibility until ReferenceGenomes are handeled by refget
+            if isinstance(element, model.ReferenceGenome):
+                source_path = Path(source_file)
+                target_path = Path(element._get("data_path"))
+                self.storage.put(source_path, target_path)
```
almutlue, line 338: I kept this, because reference genomes would not be stored otherwise. I think with the refget PR this can be removed and will be more concise/clean. So I would just keep it for now if you agree?

cmdoret: Makes sense to me, we probably want to manage them completely differently than other files in the end :+1:

modos/helpers/schema.py
```diff
@@ -111,6 +112,33 @@ def set_data_path(
...
+        else:
+            source_checksum = generate_data_checksum(source_path)
+            if source_checksum != self.model._get("data_checksum"):
+                storage.put(source_path, target_path)
+                self._set_checksum(source_checksum)
+            if source_idx:
+                storage.put(source_idx, target_idx)
+
+
 def generate_data_checksum(file_obj: Union[Path, str]) -> str:
```
cmdoret, line 142: **nitpick:** Unless there's a specific reason not to, this function would be more reusable if it took a file-like object as input (as the arg name suggests). Something like:

```python
def compute_checksum(file_obj: io.RawIOBase) -> str:
    ...

compute_checksum(open(path, 'rb'))
```

(thinking about things which don't have a path like remote files, network streams, pipes, etc)

modos/api.py
```diff
@@ -394,23 +387,12 @@ def update_element(
...
-            new_idx = get_index(new_path)
-            old_path = Path(attr_dict.get("data_path"))
-            old_idx = get_index(old_path)
-
-            if new_path != old_path:
-                self.storage.move(old_path, new_path)
-                if old_idx:
-                    self.storage.move(old_idx, new_idx)
+            new_data = DataElement(new)
+            old_path = attr_dict.get("data_path")
```
cmdoret, line 391:
```suggestion
            old_path = attr_dict.get("data_path")
            # path changed in metadata -> move file inside modos
```

suggestion: high level explanation of why we do this

modos/api.py
```diff
@@ -394,23 +387,12 @@ def update_element(
...
+            if new_data.model._get("data_path") != old_path:
+                new_data.process_and_store(self.storage, Path(old_path))
             if source_file:
```
cmdoret, line 394:
```suggestion
            # new input file provided -> replace file content
            if source_file:
```

**suggestion:** same as above

modos/helpers/schema.py
almutlue: Very good point! Thanks a lot. I updated the function accordingly.

cmdoret (approved):
LGTM! :rocket:

PR #119, opened 2025-02-27
docs: add design-notes for refget implementation

almutlue:
Design notes and background to add refget to modos

#### Open questions:
- Feedback to the sketched solutions
- Alternative solutions?
- Do I miss functionalities for modos-api integration?

### DIFF
--- a/docs/refget-design.md
+++ b/docs/refget-design.md
@@ -1,0 +1,83 @@
+# Design notes: Modos x Refget
+
+
+## Goal
+Use a [refget](http://samtools.github.io/hts-specs/refget.html) api in modos to handle reference sequences in an unambigous and standardized way.
+
+## Considerations
+1. GA4GH published a [Refget Sequence Collections](https://ga4gh.github.io/refget/seqcol/) specification (v1 approved 2025) to:
+    *   allow retrival of sequence collections (usually __reference
+ genomes__)
+    *   specify __sequence collection hashes__, that are composed from the __sequence hashes__ and __sequence attributes__ (names, length,?)(see [Seqcol standard](#seqcol-standard))
+    *   assess compatability of sequence collections 
+        *   includes coordinate systems based on sequence attributes
+2. A pangenomes standard is planned
+3. CRAM uses reference sequence ids (not sequence collections)
+
+### The Refget umbrella
+![refget-umbrella](https://ga4gh.github.io/refget/img/refget-umbrella.svg)
+
+### Seqcol standard
+![seqcol_abstract_simple](https://ga4gh.github.io/refget/img/seqcol_abstract_simple.svg)
+
+## Existing tools
+1. [refget-cloud](https://github.com/ga4gh/refget-cloud): Python-based Refget v1 web service compatible with s3 storage
+    - :heavy_check_mark: small and simple tool
+    - :heavy_check_mark: part of the official ga4gh github organisation
+    - :heavy_minus_sign: not sure how much this is maintained (last commit 5 years ago)
+    - :heavy_minus_sign: associated tooling in perl ([ena-processor](https://github.com/andrewyatz/ena-refget-processor))
+    - :heavy_minus_sign: minimalistic documentation
+2. [refgenie](https://refgenie.org/refgenie): Software suite to manage storage, access, and transfer of reference genome resources. Main usage to download pre-build genomic resources or build "assets" for custom assemblies. Based on 2 main tools: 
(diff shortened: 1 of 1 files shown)


docs/refget-design.md
```diff
@@ -0,0 +1,83 @@
...
+---
+
+## Solutions
+
+### 1. Customized Refget-cloud server
+
+Use [refget-cloud](https://github.com/ga4gh/refget-cloud) to deploy a refget server, that allows upload and retrival of reference sequences as part of modos.
+
+#### Requirements
+1. Add and adapt refget-cloud deployment to modos-deployment (inital setup [here](https://github.com/sdsc-ordes/modos-api/tree/feat/deploy_refget))
```
cmdoret, line 50: awesome that you already started this! refget-cloud seems like a good starting point :+1:

docs/refget-design.md
```diff
@@ -0,0 +1,83 @@
...
+
+
+## Goal
+Use a [refget](http://samtools.github.io/hts-specs/refget.html) api in modos to handle reference sequences in an unambigous and standardized way.
+
+## Considerations
+1. GA4GH published a [Refget Sequence Collections](https://ga4gh.github.io/refget/seqcol/) specification (v1 approved 2025) to:
+    *   allow retrival of sequence collections (usually __reference
+ genomes__)
+    *   specify __sequence collection hashes__, that are composed from the __sequence hashes__ and __sequence attributes__ (names, length,?)(see [Seqcol standard](#seqcol-standard))
```
cmdoret, line 11: sequence collections would nicely map to the ReferenceGenome concept we have, but I suggest we ignore it for now to keep it simple

docs/refget-design.md
```diff
@@ -0,0 +1,83 @@
...
+3. CRAM uses reference sequence ids (not sequence collections)
+
+### The Refget umbrella
+![refget-umbrella](https://ga4gh.github.io/refget/img/refget-umbrella.svg)
+
+### Seqcol standard
+![seqcol_abstract_simple](https://ga4gh.github.io/refget/img/seqcol_abstract_simple.svg)
+
+## Existing tools
+1. [refget-cloud](https://github.com/ga4gh/refget-cloud): Python-based Refget v1 web service compatible with s3 storage
```
cmdoret, line 24: Am I understanding correctly that it works similarly to htsget in that it does not serve the actual files, but crafts http requests and forwards them to the client?

docs/refget-design.md
```diff
@@ -0,0 +1,83 @@
...
+    - :heavy_check_mark: small and simple tool
+    - :heavy_check_mark: part of the official ga4gh github organisation
+    - :heavy_minus_sign: not sure how much this is maintained (last commit 5 years ago)
+    - :heavy_minus_sign: associated tooling in perl ([ena-processor](https://github.com/andrewyatz/ena-refget-processor))
```
cmdoret, line 28: maybe we could reimplement that? from the [algorithm outline](https://github.com/andrewyatz/ena-refget-processor/blob/7242d9c1c4cf4159d04dd6b4ca48b55977992669/docs/algorithm_outline.md?plain=1#L3), it seems that it's downloading genbank+ csv and extracting to json (and fasta?)

docs/refget-design.md
```diff
@@ -0,0 +1,83 @@
...
+    - :heavy_minus_sign: code potentially hard to extend/customize
+
+
+---
+
+## Solutions
+
+### 1. Customized Refget-cloud server
+
+Use [refget-cloud](https://github.com/ga4gh/refget-cloud) to deploy a refget server, that allows upload and retrival of reference sequences as part of modos.
```
cmdoret, line 47: I would go with this one just because it seems less complex and has fewer dependencies. This route will let us experiment with refget faster and know if it's a good fit for modos. If needed we can always change the underlying server later, as long as it follows the refget protocol, the client should be fine.

cmdoret (commented):
Thanks for the detailed writeup and context! :+1: 

Regarding your questions:

> Feedback to the sketched solutions
See comments, I agree with solution 1

> Alternative solutions?
If refget-cloud falls short for our use-case, we can extend it or reimplement it.

> Do I miss functionalities for modos-api integration?
The part that's a bit unclear to me is the upload: if someone wanted to add a new reference genome, how would they do it? Shold we even allow for that? I assume this would require more than a fasta file, and maybe some conversion work on the server side?

cmdoret:
Decided for refget cloud, implemented and documented in #121

PR #121, opened 2025-03-10
feat(deploy): setup refget service

almutlue:
### Aim: 
Deploy [refget](https://samtools.github.io/hts-specs/refget.html) service to give hash-identifier based access to reference sequences as part of modos deployment

#### Deployment:
- add a new public bucket (genome-references) to minio deployment
- deploy [refget-cloud](https://github.com/ga4gh/refget-cloud) with the genome-references bucket as source

#### Python api
- generate sequence hashes via [gtars](https://github.com/databio/gtars)
- build minimal metadata object as described in refget specs
- upload reference sequences, their metadata objects and alternative ids to s3 bucket with required `Content-Type` and `website-rdirect-location` header (see [refget-cloud s3 docs](https://github.com/ga4gh/refget-cloud/blob/master/docs/guides/DataSourceAwsS3.md); this replaces refget-compliant sequence upload via [refget-loader](https://github.com/ga4gh/refget-loader/blob/master/ga4gh/refget/loader/destinations/aws/s3/upload.py) in ga4gh pipeline)

### Testing
To test the refget deployment a reference genome can be uploaded as fasta file via:
```python
from modos.genomics.refget import RefgetStorage
from modos.storage import connect_s3
import modos.remote as remo


s3_opts = {"anon": True}
em=remo.EndpointManager(modos="http://localhost")

fasta_file="data/reference1.fa"
fs = connect_s3(em.s3, s3_opts)

rg = RefgetStorage(fs, bucket="genome-references")
rg.upload_fasta(fasta_file)
```

### Limitations
- currently the metadata object only fulfills minimal metadata requirements. This might needs to be expanded or become more flexible, when we want to be refget v.2.0.0 compliant (see next point).  
- [refget-cloud](https://github.com/ga4gh/refget-cloud) is based on refget v.1.0.0. Here, we already use `md5-checksums` as primary id and `ga4gh identifier` instead of `TRUNC512` as the default SHA-512 based hash identifier. This means this implementations is not fully compliant with neither v1.0.0 nor v2.0.0. 
Overall changes seem relatively small (mostly the use of namespaced identifiers; see [refget-appendix](https://samtools.github.io/hts-specs/refget.html#appendix-2---changes)). I would propose to make a PR to make refget-cloud v.2.0.0 compliant to solve this issue.
- [out-of-scope, priority] integration with `modos-api` and `modos-schema:ReferenceGenome` is not part of this PR. Comments on its design can be made on https://github.com/sdsc-ordes/modos-api/issues/120 
- [out-of-scope, low priority] in contrast to refget-loader we only support refget-server compliant sequence upload, but no automatic download from ENA as imlemented via [ena-processor](https://github.com/andrewyatz/ena-refget-processor). We could provide something similar in a later PR, if this really is a useful?

### DIFF
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -18,7 +18,7 @@ services:
       - REFGET_PUBLIC_URL=${REFGET_PUBLIC_URL:-http://localhost/refget}
       - FUZON_PUBLIC_URL=${FUZON_PUBLIC_URL:-http://localhost/fuzon}
       - HTSGET_PUBLIC_URL=${HTSGET_PUBLIC_URL:-http://localhost/htsget}
-      - S3_PUBLIC_URL=${S3_PUBLIC_URL:-http://localhost/s3/}
+      - S3_PUBLIC_URL=${S3_PUBLIC_URL:-http://localhost/s3}
       - FUZON_LOCAL_URL=${FUZON_LOCAL_URL:-http://fuzon:8080}
       - REFGET_LOCAL_URL=${REFGET_LOCAL_URL:-http://refget:8080}
       - HTSGET_LOCAL_URL=${HTSGET_LOCAL_URL:-http://htsget:8080}
@@ -60,7 +60,7 @@ services:
     networks:
       - modos-network
     environment:
-      - ENDPOINT=${S3_PUBLIC_URL:-http://minio:9000}
+      - ENDPOINT=${S3_PUBLIC_URL:-http://localhost/s3}
       - BUCKET=${S3_BUCKET:-modos-demo}
       - S3_ADDRESSING_STYLE=${S3_ADDRESSING_STYLE:-auto}
       - AWS_ACCESS_KEY_ID=${MINIO_ROOT_USER:-minio}
@@ -90,8 +90,8 @@ services:
     networks:
       - modos-network
     environment:
-      - ENDPOINT=${S3_PUBLIC_URL:-http://minio:9000}
-      - BUCKET=${S3_BUCKET:-genome-references}
+      - ENDPOINT=${S3_PUBLIC_URL:-http://localhost/s3}
+      - BUCKET=${REFERENCE_BUCKET:-genome-references}
 
   modos-server:
     build: ./modos-server
--- a/deploy/docker-compose.yml
+++ b/deploy/docker-compose.yml
@@ -91,7 +91,7 @@ services:
       - modos-network
     environment:
       - ENDPOINT=${S3_PUBLIC_URL:-http://localhost/s3}
-      - BUCKET=${REFERENCE_BUCKET:-genome-references}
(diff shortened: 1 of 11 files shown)


deploy/refget/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+EXPOSE 8080
+
+# gettext-base is required for envsubst (config templating)
+RUN apt update && apt install -y gettext-base libc6-dev git && apt-get clean
+
+# install uv
+COPY --from=ghcr.io/astral-sh/uv:0.6.3 /uv /uvx /bin/
+
+# install refget-cloud from github
+RUN uv pip install --system "git+https://github.com/ga4gh/refget-cloud/@c39a65acba9818414789f004cced487562012bf0"
```
cmdoret, line 12: **praise:** nice and secure :sunglasses:

deploy/docker-compose.yml
```diff
@@ -77,6 +79,20 @@ services:
...
+        - minio
+    volumes:
+      - ./refget/config.properties.template:/refget/config.properties.template:ro
+    expose:
+      - "8080"
+    networks:
+      - modos-network
+    environment:
+      - ENDPOINT=${S3_PUBLIC_URL:-http://localhost/s3}
+      - BUCKET=${REFERENCE_BUCKET:-genome-references}
```
cmdoret, line 94:
```suggestion
      - BUCKET=${REFERENCE_BUCKET:-reference-genomes}
```
**suggestion(non-blocking):** in case bucket renaming was applied above.

cmdoret (commented):
A few minor nitpicks, but this looks like a very good start: One question: all the refget stuff is computed on client side in this implementation, and all server interactions are with s3.

(When) do we need to query the refget server from modos client?

Also tests fail on my side (gonna debug in a bit)

almutlue:
Thanks a lot for the review and comments @cmdoret!
IIUC we might need/want a way to list all existing reference genomes and check if a reference sequence already exists (by hash comparison?) or did you refer to something else?

cmdoret:
I was just surprised that there is no direct interaction between the modos client and the refget server, but we may not need it.

This discussion is probably for another PR, but here are the flows as I understand them:

* modos client uploads new reference
  1. sequences+digests uploaded to reference bucket (if not already present)
  2. ReferenceGenome+ReferenceSequence json uploaded to modos bucket
    +  json links to reference sequence uri via `ReferenceSequence->data_path`
* external software (e.g. samtools) reads cram file
  1. cram file header contains link to our refget sequence uri
  2. samtools queries this uri
  4. refget server crafts s3 request from uri
  5. samtools queries crafted s3 request

Maybe at some point we want to provide the option to materialize or even stream a fasta file. In that case, there would indeed be direct interaction between client and refget.

cmdoret (approved):
:zap: :dna: :zap:

PR #123, opened 2025-03-21
feat(cli): update prompts for removal of missing config elements

almutlue:
### Aim: 
Add force flag to `modos update`. See https://github.com/sdsc-ordes/modos-api/issues/114

### Changes:
Replace `--no-remove` with `--force` flag in `modos update`. Instead of explicitly allowing to __not__ remove missing elements, this PR implements the following logic, if an existing element is missing from `config_file`:
 - prompt for explicit deletion by default 
 - add --force flag to skip prompts

### DIFF
--- a/modos/cli.py
+++ b/modos/cli.py
@@ -22,7 +22,7 @@ from modos.codes import get_slot_matcher, SLOT_TERMINOLOGIES
 from modos.helpers.schema import UserElementType
 from modos.genomics.htsget import HtsgetConnection
 from modos.genomics.region import Region
-from modos.io import parse_instance
+from modos.io import parse_instance, parse_attributes
 from modos.prompt import SlotPrompter
 from modos.remote import EndpointManager
 from modos.prompt import SlotPrompter, fuzzy_complete
@@ -389,12 +389,12 @@ def update(
             help="File defining the updated modo. The file must be in json or yaml format.",
         ),
     ],
-    no_remove: Annotated[
+    force: Annotated[
         bool,
         typer.Option(
-            "--no-remove",
-            "-n",
-            help="Do not remove elements that are missing in the config_file.",
+            "--force",
+            "-f",
+            help="Force deletion of elements that are missing in the config_file.",
         ),
     ] = False,
 ):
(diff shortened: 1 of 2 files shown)


modos/cli.py
```diff
@@ -389,25 +389,50 @@ def update(
...
+            no_remove=True,
+        )
+        modo_id = _.zarr["/"].attrs["id"]
+        meta_ids = {Path(id).name: id for id in _.metadata.keys()}
+        old_ids = [
+            id for id in meta_ids.keys() if id not in new_ids and id != modo_id
+        ]
+        for old_id in old_ids:
+            delete = typer.confirm(
+                f"Could not find element '{old_id}' in config_file.\n Shall {old_id} be deleted?"
```
cmdoret, line 428:
```suggestion
                f"Object contains element '{old_id}' which was not in config_file.\n Delete {old_id} ?"
```

**suggestion:** rephrase to emphasize discrepancy between object state and config.

modos/cli.py
```diff
@@ -389,25 +389,50 @@ def update(
...
+            )
+            if not delete:
+                print(
+                    f"Keep {old_id} as part of {modo_id}. Consider updating your config_file."
```
cmdoret, line 432:
```suggestion
                    f"Keeping {old_id} in of {modo_id}. Consider updating your config_file."
```
**nitpick:** phrasing

modos/cli.py
```diff
@@ -389,25 +389,50 @@ def update(
...
+    if force:
+        _ = MODO.from_file(
+            config_path=config_file,
+            object_path=object_path,
+            endpoint=endpoint,
+            no_remove=False,
+        )
+    else:
+        element_list = parse_attributes(Path(config_file))
+        new_ids = [ele["element"].get("id") for ele in element_list]
```
cmdoret, line 414: **suggestion(non-blocking):** might be less confusing to rename variable `new_ids` -> `config_ids` ?

cmdoret (approved):
Nice! provisionally approving, my suggestions are just semantic nitpicking ;)

modos/cli.py
almutlue: Thanks - good point and better name!

PR #126, opened 2025-04-25
fix(server): api compatibility for /meta and /get

cmdoret:
## Summary

This fixes server side bugs due to breaking changes in the API:
* endpoint is now passed correctly to MODO
* paths of remote modos are pre-pended with the `s3://` scheme

The only file affected by this PR is [`tools/deploy/modos-server/server.py`](https://github.com/sdsc-ordes/modos-api/pull/126/files#diff-d9ee2e2ad6e09b3311ae9d25e2821dcfc9ca6426a23b0a0c4948cc69bcb8b69c)


almutlue (approved):
Well spotted and fixed. Looks good to me and works as expected now.

PR #127, opened 2025-04-25
feat: local crypt4gh encryption

almutlue:
### Purpose:
- add download/upload functions from remote
- add c4gh decryption/encryption of genomic files

### Notes:
- `crypt4gh.encrypt()` auto generates a secretkey, if not provided. I mimicked that behavior.
- files can be encrypted for multiple users, but `modos.encrypt()` will apply the same encrytion on all files included.

### Questions:
- `modos.api.encrypt()` and `modos.api.decrypt()` feel clunky. The main reason for this is the handling of index files. They can not directly be derived from metadata and thus require some extra logic. I also think we can not always assume them to exist (?), which causes extra checks. I'm open for any suggestions to move code away from the api to keep it cleaner?    
- Do we need to include a check if the modo is local to `modo.encrypt()` and `modo.decrypt()`


modos/helpers/schema.py
```diff
@@ -132,12 +141,15 @@ def process_and_store(self, storage, source_path: Path):
             if source_idx:
                 storage.move(source_idx, target_idx)
         else:
-            source_checksum = compute_checksum(open(source_path, "rb"))
-            if source_checksum != self.model._get("data_checksum"):
-                storage.put(source_path, target_path)
+            with open(source_path, "rb") as src:
+                source_checksum = compute_checksum(src)
+            with open(source_path, "rb") as src:
```
almutlue, line 146: `src.seek(0)`

modos/api.py
```diff
@@ -562,3 +571,69 @@ def from_file(
...
+                        encrypt_file(
+                            recipient_pubkeys=recipient_pubkeys,
+                            infile=self.path / file_path,
+                            outfile=self.path / out_path,
+                            seckey_path=seckey_path,
+                            passphrase=passphrase,
+                        )
+                        self.storage.remove(self.path / file_path)
+                if self.storage.exists(encrypted_path):
+                    update_data_path(meta, encrypted_path, self.path)
```
almutlue, line 610: update DataEntity with DataElement (self.meta[...] = data.meta`)

modos/api.py
```diff
@@ -562,3 +571,69 @@ def from_file(
...
+    def encrypt(
+        self,
+        recipient_pubkeys: List[os.PathLike] | os.PathLike,
+        seckey_path: Optional[os.PathLike] = None,
+        passphrase: Optional[str] = None,
+    ):
+        """Encrypt genomic data files including index files in a modo using crypt4gh"""
+        for group in self.zarr.data.values():
+            meta = group.attrs
+            if meta["data_format"] in GenomicFileSuffix.formats:
```
almutlue, line 590: Move to DataElement
`DataElement.encrypt(delete = True)`

modos/api.py
```diff
@@ -562,3 +571,69 @@ def from_file(
         for old_id in old_ids:
             modo.remove_element(modo_ids[old_id])
         return modo
+
+    def download(self, target_dir: Path):
+        """Download the MODO to a local directory.
+        This will download all files and metadata to the target directory.
+        """
+        self.storage.download(target_dir)
```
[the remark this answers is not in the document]
almutlue: Add upload function

modos/cli.py
```diff
@@ -223,7 +242,90 @@ def enrich(
...
+        typer.Option(
+            "--secret-key",
+            "-s",
+            help="Secret key of the sender to encrypt files in the MODO.",
+        ),
+    ] = None,
+):
+    """Encrypt a local MODO."""
+    modo = MODO(object_path)
+    modo.encrypt(public_key, secret_key)
```
almutlue, line 325: Include passphrase --p -> file (open.read())

almutlue:
Finally, ready for the next round of review @cmdoret.

### Main changes:
- upload/download function in api + tests (triggered by `pytest --runslow`)
- refactor `modos.api.MODO.encryption()/decryption()` to use `DataEntity.encrypt()/decrypt()` to handle index files outside api code 
-  add encryption/decryption tests for cli.

modos/helpers/schema.py
```diff
@@ -114,30 +147,110 @@ def set_data_path(
...
+            self.model.data_format.code.text
+            in GenomicFileSuffix.list_formats()
+        )
 
     def process_and_store(self, storage, source_path: Path):
         source_idx = get_index(source_path)
-        target_path = Path(self.model._get("data_path"))
+        target_path = Path(self.model.data_path)
         target_idx = get_index(target_path)
         if storage.exists(source_path) and source_path != target_path:
```
cmdoret, line 177:
```suggestion
        # Renaming / moving existing file within modo.
        if storage.exists(source_path) and source_path != target_path:
```

**suggestion:** explain what's going on

modos/helpers/schema.py
```diff
@@ -114,30 +147,110 @@ def set_data_path(
...
             storage.move(source_path, target_path)
             if source_idx:
                 storage.move(source_idx, target_idx)
         else:
```
cmdoret, line 181:
```suggestion
        # Importing / overwriting external file into modo.
        else:
```
**suggestion:** explain what's going on

cmdoret (changes requested):
Nice work! My suggestions mostly revolve around semantics / documentation :+1:

modos/genomics/c4gh.py
```diff
@@ -0,0 +1,99 @@
...
+from pathlib import Path
+
+from crypt4gh.keys import get_public_key, get_private_key
+from crypt4gh.lib import decrypt, encrypt
+from functools import partial
+from getpass import getpass
+from nacl.public import PrivateKey
+
+
+def get_secret_key(
```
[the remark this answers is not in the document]
almutlue: I agree that the option feels a bit odd. I kept it to allow people to encrypt files without generating a secret key in advance. 
The way it is [implemented in c4gh](https://github.com/EGA-archive/crypt4gh/blob/dec3af5ac713f7fe8fbdd96d871d3abbd8c725fc/crypt4gh/cli.py#L103) is a bit convoluted:
1. Encryption requires a secret key. `crypt4gh` is authenticated encryption - AFAIK the algorithm has two main steps: 
- header encryption with asymmetric keys
- data encryption with symmetric keys. 
Symmetric keys are encrypted with a shared secret which requires the writer's private key.
2. If the writer's secret key is not provided, it will be silently generated (`generate = True`)
3. `get_secret_key` is also used for decryption with `generate = False` (see [here](https://github.com/sdsc-ordes/modos-api/blob/2ed7124c930af03211893bca5bdb4493860b1d8b/src/modos/genomics/c4gh.py#L86)) 

I kept the option, but added docstrings to `get_secret_key()` and `encrypt_file()` to clarify what the option is for and that crypt4gh is authenticated encryption.

modos/helpers/schema.py
almutlue: Agree - I think this is still hard to fully entagle. I will try a complete refactoring as part of fixing https://github.com/sdsc-ordes/modos-api/issues/129

modos/genomics/c4gh.py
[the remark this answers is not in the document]
cmdoret: OK I understand the use-case. Would it make sense to drop `generate` and still have the `Optional` seckey, such that None means a key will be generated? (ofc it would need to be explicitely stated in the docstring + maybe a output a log message saying it's generating it).

modos/genomics/c4gh.py
```diff
@@ -0,0 +1,99 @@
...
+    outfile: Path | str,
+    sender_pubkey: Optional[os.PathLike] = None,
+    passphrase: Optional[str] = None,
+):
+    seckey = get_secret_key(seckey_path, generate=False, passphrase=passphrase)
+    keys = [
+        (0, seckey, None)
+    ]  # keys = list of (method, privkey, recipient_pubkey=None)
+    if sender_pubkey:
+        sender_pubkey = os.path.expanduser(sender_pubkey)
```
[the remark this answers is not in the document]
cmdoret: My bad, indeed you're right we use it! nevermind :sweat_smile:

tests/test_api_remote.py
```diff
@@ -124,3 +124,22 @@ def test_update_source_file_and_data_path(remote_modo):
...
     assert remote_modo.storage.exists("demo1.cram")
     assert not remote_modo.storage.exists("demo2.cram")
+
+
+# Upload/download entire modo
+@pytest.mark.slow
+def test_upload_modo(setup, test_modo):
+    minio_client = setup["minio"].get_client()
+    minio_endpoint = setup["minio"].get_config()["endpoint"]
+    minio_creds = {"secret": "minioadmin", "key": "minioadmin"}
```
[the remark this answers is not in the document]
cmdoret: OK Might be a bug, we can figure this out separately, not a big deal

cmdoret (approved):
Feel free to apply the `generate` suggestion or not, besides that it all LGTM :+1:

PR #135, opened 2025-05-20
ci(docker): fix version parsing  on release

cmdoret:
* Fetch version from `__init__.py` instead of `pyproject.toml`
* Allow pushing docker images with version tags manually using workflow dispatch (in addition to on release)


almutlue (approved):
Looks very good to me! Thanks a lot!

PR #138, opened 2025-05-23
fix(api): sequential encryption

almutlue:
## Changes

- check if files are encrypted through file header bytes.
- handles file suffixes unambigously (`add` or `remove` suffix instead of `toggle`).
- (decryption): raise error if c4gh encrypted, but no `c4gh`suffix. This should not happen. Modos and c4gh libs add `.c4gh` as suffix, but this is no strict requirement.

## Limitations
- only checks for crypt4gh encryption. I think that is reasonable as we only support c4gh encryption

### DIFF
--- a/src/modos/genomics/formats.py
+++ b/src/modos/genomics/formats.py
@@ -7,6 +7,8 @@ import pysam
 
 from modos.genomics.region import Region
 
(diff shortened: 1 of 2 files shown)


cmdoret (approved):
LGTM :lock:

PR #140, opened 2025-05-23
fix(cli): syntax when reading passphrase from file

cmdoret:
Fixes a crash when passphrase is read from a file in `modos decrypt`


almutlue (approved):
Somehow I missed this. Sorry for the delay and thanks a lot for the fix. Tested and works!

PR #144, opened 2025-06-03
refactor(nix): devenv-based python

cmdoret:
Provides a fully encapsulated python environment using a nix devenv configuration.
This avoids relying on uv's downloaded python which will link to system C libraries. With this setup, every dependency is in the nix store.


tools/nix/flake.nix
```diff
@@ -11,44 +11,49 @@
...
+      let
         # Import nixpkgs and load it into pkgs.
         pkgs = import nixpkgs {
           inherit system;
         };
 
-        # Things needed at build-time.
-        packagesBasic = with pkgs; [
+        # python environment defined separately
+        pythonModules = import ./python-uv.nix {
```
gabyx, line 33: I would call that file rather python-modules or so, because it can be mistaken that it returns a python derivation

tools/nix/flake.nix
```diff
@@ -11,44 +11,49 @@
...
           bash
           coreutils
           curl
           findutils
           git
           git-cliff
           just
-          pyright
-          uv
           zsh
```
gabyx, line 47: Move that maybe next to bash :)

gabyx (approved):

PR #145, opened 2025-06-13
ci(docker): fix release tag

cmdoret:
* Fixes tag extraction in docker publish CI workflow
* Bumps the tag of the base layer for the server image


cmdoret:
Good point, I've opened https://github.com/sdsc-ordes/modos-api/issues/149 to track it

PR #146, opened 2025-06-13
chore(server): bump fuzon to 0.4.0

cmdoret:


almutlue:
0.4.0 works for me as it is supposed to work :+1:

PR #148, opened 2025-06-15
fix(storage): transfer metadata

cmdoret:
The `LocalStorage` implementation of `list` was skipping the `data.zarr` dir. This prevented `Storage.transfer()` from transferring metadata files when running on a local souce modo.


almutlue (approved):
Well spotted bug :open_mouth:. I also added a small fix for ``modos.list_files() ` resp. modos show -f`. I guess that is why `data.zarr` was excluded in the past? 
I tested it local and remote, but please a quick check before merging :eyes:

cmdoret:
Seems to work :+1:

PR #150, opened 2025-06-16
style: embed rich in dependencies

cmdoret:
Adds rich as an extra dependency of typer to ensure consistent rendering of CLI.


almutlue (approved):
styling is always nice :smile:

PR #155, opened 2025-06-25
deploy: add compose profiles

almutlue:
### Goal:
Add compose profiles to enable skipping of minio deployment

### Details:
- adds compose profile local to minio service
- adapts `depends_on` of other services to only apply, if minio is available
- adapt just file to run deploy for local profile

### Comments:
I also replaced the `--chmod` option in the refget Dockerfile. so that we can also build with the docker legacy builder (this should fix: https://github.com/sdsc-ordes/modos-fega/pull/4#discussion_r2166258108). We can also revert this, if you prefer a different solution?


cmdoret (approved):
I like this solution as well :+1: LGTM

PR #157, opened 2025-07-04
fix(deploy): nonroot container users

cmdoret:
Set nonroot users in all service containers.


LuanaMartelli:
changes look good :+1: 
If I may add while we are at it, trivy scans also pointed out:
- use COPY instead of ADD instructions 
- when you `apt-get`, use the flag `--no-install-recommends` if possible to minimize the size of the image
- optional: add health checks instructions (but maybe this would be more relevant in a docker-compose or in k8s config)

cmdoret:
Thanks @LuanaMartelli ! We now use COPY and --no-install-recommends.
We previously discussed health checks and plan to implement that at the docker compose level.

almutlue (approved):
LGTM - thanks a lot, both of you!

PR #179, opened 2025-09-30
feat(cli): pass s3 credentials via env

cmdoret:
Adds support for s3 credentials through the command line interface via environment variables.

## Changes

The default behavious in modos until now was to pass `{"anon": True}` to the s3fs client.
This PR chnages the default to False so that it automatically scans standard AWS S3 environment variables.

Anonymous access is still possible through a new `--anon` toplevel flag on the CLI.


almutlue (approved):
Tested and works for read and write operations :100:

almutlue (approved):
Great, looks really good :+1: ! I have to admit that I forgot that the EndpointManager does a server call by default :eyes:

PR #192, opened 2025-12-15
fix(deploy): use build arg to allow change of server base image

almutlue:
## Summary

This should allow to build the server image with a different base image, e.g.:
```
docker build --build-arg MODOS_API_IMAGE=ghcr.io/sdsc-ordes/modos-api:latest
```

This should allow using the `latest` in dev deployments.


cmdoret:
Clean and simple, works fine on my side :woman_cook:

cmdoret (approved):

PR #195, opened 2025-12-20
feat(deploy): kms integration

cmdoret:
## Goal

Adds a key management service (kms) as part of the deployment compose stack, to automate s3 key provisioning.

## Context

The service is only enabled if compose is started with `--profile auth`, and is intended for the case where access to services is gated by an identity provider.

Instead of the modos client having a fixed S3 access key, this allows `modos remote login` to dynamically requests a temporary S3 access key to the kms which will have the same life time and permissions as the JWT obtained from the identity provider.

## Changes

* [x] add kms service to compose stack
* [x] bump garage v1 -> v2 to support API
* [x] enable garage API in configuration
* [x] add environment variables for S3 API and Authentik token
* [x] implement kms flow in `remote login`, handling expired s3 key
* [x] source cached credentials in cli and pass to storage through API
* [x] document flow
* [x] rebase on #194 to test with token containing relevant scopes

```mermaid
sequenceDiagram
    participant C as Client
    participant IdP as Identity Provider
    participant KMS as Key Management Service
    participant S3 as S3 Server

    C->>IdP: Login (device code flow)
    activate IdP
    IdP-->>C: JWT token
    deactivate IdP

    C->>KMS: Request S3 access key (with JWT)
    activate KMS
    KMS->>KMS: Validate JWT
    KMS->>S3: Request new access key (match JWT lifetime & permissions)
    activate S3
    S3-->>KMS: Return S3 access key
    deactivate S3
    KMS-->>C: Return S3 access key
    deactivate KMS

    C->>C: Persist S3 access key
```

## Limitations

Currently, the kms only supports garage (v2+) for the S3 server and authentik as the identity provider.

### DIFF
--- a/tools/deploy/kms/src/kms.py
+++ b/tools/deploy/kms/src/kms.py
@@ -84,7 +84,7 @@ def decode_token(token: str) -> dict[str, str | list[str]]:
             algorithms="HS256",
             audience=client_id,
             verify_audience=True,
-            require = ["exp", "iat", "iss", "aud"],
+            require=["exp", "iat", "iss", "aud"],
         )
     except jwt.ExpiredSignatureError:
         raise HTTPException(status_code=401, detail="Token has expired")


cmdoret:
> Current status:
> 
>     * kms receives jwt
> 
>     * token successfully decoded
> 
>     * garage does not accept the request body when requesting new key
> 
> 
> Error message from garage API:
> 
> ```
> garage-1        | 2025-12-20T18:46:25.671629Z  INFO garage_api_common::generic_server: Response: error 400 Ba
> d Request, Bad request: Invalid JSON: premature end of input at line 1 column 84
> kms-1           | PAYLOAD
>  {'name': '56dae52a-c098-41d9-93b7-2459c0241046', 'expiration': '2025-12-20T18:53:32'}
> kms-1           | RESPONSE
> kms-1           | {'code': 'InvalidRequest', 'message': 'Bad request: Invalid JSON: premature end of input at
>  line 1 column 84', 'region': 'ch', 'path': '/v2/CreateKey'}
> kms-1           | INFO:     172.19.0.8:57458 - "POST /create HTTP/1.1" 500 Internal Server Error
> ```

fixed by passing all fields of the request (all are mandatory, but the error message is not clear)

tools/deploy/kms/src/kms.py
```diff
@@ -0,0 +1,198 @@
...
+    # NOTE: only symmetric HS256 is supported with authentik proxy providers
+    # hence the need for client_secret
+    client_id, client_secret = get_client_credentials()
+    try:
+        decoded = jwt.decode(
+            token,
+            client_secret,
+            algorithms="HS256",
+            audience=client_id,
+            verify_audience=True,
```
almutlue:
```suggestion
            verify_audience=True,
            require = ["exp", "iat", "iss", "aud"],
```
__todo__: If not explicitly required, expiration will only be validated if present (see the warning here: https://pyjwt.readthedocs.io/en/stable/api.html#jwt.decode).

almutlue (changes requested):
Really nice work :rocket: . I just have a few minor comments mostly regarding the jwt decoding.

When testing `modos remote upload` and `download` work perfectly :100:.
`modos remote stream` returned a KeyError (see below). Does it work on your side? Could it be related to the use of subdirectories within the bucket (e.g `modos-demo/submitters/ex`)
I would consider this a separate PR, if there is no quick fix.

```sh
KeyError: "No htsget urls found in ticket: {'htsget': {'error': 'NotFound', 'message': 'aws error: unhandled error, with key:
`submitters/ex1/demo1.cram`'}}"
```

I also noticed that the list endpoint of modos server returns the bucket level (see below), but misses modos stored in subdirectories (groups). This should probably also be a separate PR as it is just a side-effect of using groups/subdirectories.

```sh
modos --endpoint http://localhost remote list
s3://modos-demo/submitters
```

almutlue (approved):
LGTM. Thanks a lot!

PR #212, opened 2026-06-12
fix(htsget): overlap behavior for enclosed regions

cmdoret:
## Context

`modos.genomics.region.Region.overlap()` does not report overlaps for completely enclosed regions.
This is illustrated by the failing test added in e0d33dc
```

________________________ test_overlap_contained_region _________________________

    def test_overlap_contained_region():
        """A region fully contained in another overlaps it, both ways."""
        outer = Region(chrom="chr1", start=10, end=100)
        inner = Region(chrom="chr1", start=40, end=50)
        assert outer.overlaps(inner)
>       assert inner.overlaps(outer)
E       AssertionError: assert False
E        +  where False = overlaps(Region(chrom='chr1', start=10, end=100))
E        +    where overlaps = Region(chrom='chr1', start=40, end=50).overlaps

```

## Changes

* overlap logic now correctly identifies contained intervals as overlaps.


almutlue (approved):
Looks good!

PR #214, opened 2026-06-12
feat(htsget): c4gh-encrypted streams

cmdoret:
## Changes
* Support for crypt4gh-encrypted "streams" to the htsget client.
* Update documentation with new parameters.
* Stream decryption tests.

More details in the [specification document](https://github.com/sdsc-ordes/modos-api/blob/7efa7892c5fb336fa8ca4748d9ef063c2877aff9/docs/specs/2026-06-12-encrypted-htsget-streams-design.md).

## Limitations :warning:

* The htsget spec does not define how c4gh should be handled. This PR follows the htsget-rs implementation of c4gh, meaning it might not work with other server implementations.
* This uses [`crypt4gh.lib.decrypt`](https://github.com/EGA-archive/crypt4gh/blob/7e1d4053c8b04995f32081fa564f014242a214eb/crypt4gh/lib.py#L325) which decrypts the whole file -> we are storing the whole decrypted file on disk -> not proper streaming. crypt4gh has other functions to decrypt specific parts, but I suggest to move to these once we have proper streaming support implemented e.g. after #215 .
* I did not test this end-to-end with a real server.

> [!NOTE]
> Tests are not fully human and a bit convoluted overkill. I left them, but I would suggest to drop some of them and replace with end-to-end testing if that makes sense.


src/modos/cli/remote.py
```diff
@@ -169,6 +170,23 @@ def stream(
             help="Restrict stream to genomic region (chr:start-end).",
         ),
     ] = None,
+    secret_key_path: Annotated[
+        Optional[Path],
```
almutlue: __nitpick__: we switched to a `|` notation
```suggestion
        Path | None,
```

src/modos/cli/remote.py
```diff
@@ -169,6 +170,23 @@ def stream(
...
+        typer.Option(
+            "--secret-key",
+            "-s",
+            help="Secret key to decrypt an encrypted stream. Its public "
+            "key is sent to the htsget server.",
+        ),
+    ] = None,
+    passphrase: Annotated[
+        Optional[Path],
```
almutlue:
```suggestion
        Path | None,
```

src/modos/genomics/c4gh.py
almutlue, line 97: __Question__: This is out of scope of this PR, but I wonder if we should by default include the senders public key here to make sure access is not lost upon encryption (by default we delete the input)?  `derive_public_key()` makes this a one-liner. 

```suggestion
recipient_list.append((0, seckey, derive_public_key(seckey)))
```

almutlue (approved):
Nice, looks good and I agree with the scope.

